PR #19, opened 2018-03-23
Fix VS update to 15.6.1 case

### LINKED ISSUE #17: `QueryVisualStudioInstances()` not work with visual studio 15.6.1
`MSBuildLocator.QueryVisualStudioInstances()` returns 0 with the latest VS 2017 15.6.1

vicancy:
Fix #17

### DIFF
--- a/src/MSBuildLocator/VisualStudioLocationHelper.cs
+++ b/src/MSBuildLocator/VisualStudioLocationHelper.cs
@@ -51,7 +51,8 @@ namespace Microsoft.Build.Locator
                         continue;
 
                     // If the install was complete and a valid version, consider it.
-                    if (state == InstanceState.Complete)
+                    if (state == InstanceState.Complete ||
+                        (state.HasFlag(InstanceState.Registered) && state.HasFlag(InstanceState.NoRebootRequired)))
                         validInstances.Add(new VisualStudioInstance(
                             instance.GetDisplayName(),
                             instance.GetInstallationPath(),


vicancy:
Thanks @heaths, could you share where to get the `formatted error log`? Is it `ISetupInstance2.GetErrors()`? It only contains `GetFailedPackages` however no error messages in there.

heaths:
QI on `ISetupErrorState` returned from `GetErrors` for for [ISetupErrorState2](https://docs.microsoft.com/en-us/dotnet/api/microsoft.visualstudio.setup.configuration.isetuperrorstate2?view=visualstudiosdk-2017) and you can call `GetErrorLogFilePath`.

vicancy:
Thanks @heaths It looks working although the log does not exist in my local machine, probably got cleaned up.. Is there any additional error code to indicate if the failure is a fatal one or not?

Hi @rainersigwald , what's your opinion? Looks like I ran into this state because some packages failed to install, It CAN be packages with minor features that does not affect VS running, it can also be packages that affects the core feature of VS. In my case, it is some minor feature that fails. Do you think it is more accurate to return the instance if it is `Registered` and meanwhile return the error message?

heaths:
If the failure if vital failure (failure in a vital package, for which there are few) the `Registered` state should not be present.

heaths:
The path - apart from having the version number in it despite now being part of Visual Studio - is fixed. It hasn't changed structure in many years. Did you take a look at my sample on the vswhere wiki? You could codify that.

rainersigwald:
@heaths I meant that this assembly/repo exists to make it easier to use this than to keep manually constructing environment variables hard-coded to `C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\MSBuild\15.0\Bin` that screw up all your hard work on the setup side. Since that is extremely permissive with respect to VS install state, I think we should be too--it's still strictly better this way.

heaths:
I wasn't arguing against this repo. I'm saying that I don't think constructing the path based on a fixed directory structure is naïve. I was referencing [this example](https://github.com/Microsoft/vswhere/wiki/Find-MSBuild) as an approach to constructing paths, though I realize (now that I'm typing from my phone) it wasn't the sample I was thinking of. I wrote one once (maybe it was in an email) that basically took the major version component of the `InstallationVersion` of the found instance and used it in the path, e.g. "MSBuild\<Major>.0\Bin\MSBuild.exe". MSBuild has used that directory structure for many years and I doubt would change based on internal conversation on this very topic.

PR #24, opened 2018-04-05
Check PublicKeyToken before resolving an assembly

AndyGerlicher:
* If there is a request for one of our assemblies with another key do
not handle the AssemblyResolve event to let other resolvers have a
chance.

### DIFF
--- a/src/MSBuildLocator/MSBuildLocator.cs
+++ b/src/MSBuildLocator/MSBuildLocator.cs
@@ -6,11 +6,14 @@ using System.Collections.Generic;
 using System.IO;
 using System.Linq;
 using System.Reflection;
+using System.Text;
 
 namespace Microsoft.Build.Locator
 {
     public static class MSBuildLocator
     {
+        private const string MSBuildPublicKeyToken = "b03f5f7f11d50a3a";
+
         private static readonly string[] s_msBuildAssemblies =
         {
             "Microsoft.Build", "Microsoft.Build.Framework", "Microsoft.Build.Tasks.Core",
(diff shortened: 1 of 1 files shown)


AndyGerlicher:
@rainersigwald test this please

src/MSBuildLocator/MSBuildLocator.cs
```diff
@@ -101,7 +104,24 @@ private static bool IsMSBuildAssembly(Assembly assembly)
...
+            var publicKeyToken = assemblyName.GetPublicKeyToken();
+            if (publicKeyToken == null || publicKeyToken.Length == 0)
+            {
+                return false;
+            }
+
+            var sb = new StringBuilder();
+            foreach (var b in publicKeyToken)
+            {
+                sb.Append($"{b:x2}");
```
rainersigwald: ew.

src/MSBuildLocator/MSBuildLocator.cs
```diff
@@ -101,7 +104,24 @@ private static bool IsMSBuildAssembly(Assembly assembly)
...
+                return false;
+            }
+
+            var publicKeyToken = assemblyName.GetPublicKeyToken();
+            if (publicKeyToken == null || publicKeyToken.Length == 0)
+            {
+                return false;
+            }
+
+            var sb = new StringBuilder();
```
rainersigwald:
```C#
new StringBuilder(MSBuildPublicKeyToken.Length);
```

to avoid allocations. Though I guess there'll be garbage strings for each byte later so it probably doesn't matter.

PR #32, opened 2018-04-18
Report better version for VS preview consoles

rainersigwald:
For prereleases, the Visual Studio command prompt sets the environment
variable `VSCMD_VER` to a value like `15.7.0-pre.3.0`, which is not
parseable by System.Version.

This could be handled by using a version parser that can handle this
SemVer 2 version as well as Windows 4-part versions, but that would be
another dependency, so I just special-cased the problematic dash.

### DIFF
--- a/src/MSBuildLocator/MSBuildLocator.cs
+++ b/src/MSBuildLocator/MSBuildLocator.cs
@@ -143,6 +143,12 @@ namespace Microsoft.Build.Locator
                 Version version;
                 Version.TryParse(versionString, out version);
 
+                if (version == null && versionString.Contains('-'))
+                {
+                    versionString = versionString.Substring(0, versionString.IndexOf('-'));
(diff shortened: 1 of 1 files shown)


rainersigwald:
example before:

```
Sample MSBuild Builder App 1.0.15+g18a79eb502.

1) DEVCONSOLE - 15.0 (Recommended!)
2) Visual Studio Enterprise 2017 - 15.7.27617.1
3) Visual Studio Enterprise 2017 - 15.0.26228.4
```

after:

```
Sample MSBuild Builder App 1.0.14+g459ba6b7dd.

1) DEVCONSOLE - 15.7.0 (Recommended!)
2) Visual Studio Enterprise 2017 - 15.7.27617.1
3) Visual Studio Enterprise 2017 - 15.0.26228.4
```

PR #33, opened 2018-04-30
Support for locating .NET Core instances

jeffkl:
Copy code to locate dotnet SDK

NET46 version of the assembly can only locate full framework MSBuild.  NETSTANDARD version of the assembly can only locate .NET Core MSBuild.

Fixed logic bug in filtering results
Added some unit tests
Updated sample app


jeffkl:
Sample output:

```
D:\MSBuildLocator>dotnet D:\MSBuildLocator\samples\BuilderApp\bin\Debug\netcoreapp2.1\BuilderApp.dll D:\MSBuildLocator\samples\BuilderApp\test.proj
Sample MSBuild Builder App 1.0.16+g8664658d65.

1) .NET Core SDK - 2.1.300

Select an instance of MSBuild:
1
Using VS Instance: .NET Core SDK - 2.1.300


BuildApp running using MSBuild version 15.5.180.51428
C:\Users\jeffkl\.nuget\packages\microsoft.build\15.5.180\lib\netstandard1.5

Build started.
Project "test.proj" (default targets):
Building with tools version "15.0".
Project file contains ToolsVersion="". This toolset may be unknown or missing, in which case you may be able to resolve this by installing the appropriate version of MSBuild, or the build may have been forced to a particular ToolsVersion for policy reasons. Treating the project as if it had ToolsVersion="15.0". For more information, please see http://go.microsoft.com/fwlink/?LinkId=293424.
Target "Build" in project "D:\MSBuildLocator\samples\BuilderApp\test.proj" (entry point):
Task "Error" skipped, due to false condition; (!$(MSBuildExtensionsPath.StartsWith('$(ProgramFiles(x86))'))) was evaluated as (!True).
Done building target "Build" in project "test.proj".
Done building project "test.proj".
Build succeeded.

Build result: True
```

src/MSBuildLocator/DiscoveryType.cs
```diff
@@ -20,6 +20,11 @@ public enum DiscoveryType
         /// <summary>
         ///     Discovery via Visual Studio Setup API.
         /// </summary>
-        VisualStudioSetup = 2
+        VisualStudioSetup = 2,
+
+        /// <summary>
+        /// Discovery via dotnet (hostfxr).
+        /// </summary>
+        DotNetSdk = 4
```
DustinCampbell: Why is this a flags enum? In practice, can't it only ever be one of these values?

src/MSBuildLocator/DotNetSdkLocationHelper.cs
```diff
@@ -0,0 +1,124 @@
...
+{
+
+    using System;
+    using System.IO;
+    using System.Runtime.InteropServices;
+    using System.Text;
+
+    internal static class DotNetSdkLocationHelper
+    {
+        public static VisualStudioInstance GetInstance(string workingDir)
```
DustinCampbell: It feels like `VisualStudioInstance` was a kind of poor name choice since Visual Studio really isn't involved for the .NET Core SDK case. :smile:

src/MSBuildLocator/DotNetSdkLocationHelper.cs
```diff
@@ -0,0 +1,124 @@
...
+            }
+
+            if (!File.Exists(Path.Combine(dotNetSdkPath, "Microsoft.Build.dll")))
+            {
+                return null;
+            }
+
+            FileInfo dotnetAssemblyPath = new FileInfo(Path.Combine(dotNetSdkPath, "dotnet.dll"));
+
+            if (!dotnetAssemblyPath.Exists)
```
DustinCampbell: For my education, what's the reason for using `File.Exists` above but `FileInfo.Exists` here?

src/MSBuildLocator/DotNetSdkLocationHelper.cs
```diff
@@ -0,0 +1,124 @@
...
+
+        private static string hostfxr_resolve_sdk(string workingDir)
+        {
+            Process process = Process.GetCurrentProcess();
+
+            string dotNetExePath = process?.MainModule?.FileName;
+
+            if (String.IsNullOrWhiteSpace(dotNetExePath))
+            {
+                return null;
```
DustinCampbell: Should we throw here or provide the user with some explanation for why we couldn't locate MSBuild? If it fails, it feels like this is going to be a head-scratcher. There are a lot of failure conditions here that we could provide (somehow) more information to the user about.

src/MSBuildLocator/DotNetSdkLocationHelper.cs
```diff
@@ -0,0 +1,124 @@
...
+            FileVersionInfo fileVersionInfo = FileVersionInfo.GetVersionInfo(dotnetAssemblyPath.FullName);
+
+            return new VisualStudioInstance(
+                name: ".NET Core SDK",
+                path: dotNetSdkPath,
+                version: new Version(fileVersionInfo.FileMajorPart, fileVersionInfo.FileMinorPart, fileVersionInfo.FileBuildPart),
+                discoveryType: DiscoveryType.DotNetSdk);
+        }
+
+        private static string hostfxr_resolve_sdk(string workingDir)
```
DustinCampbell: You're missing this check to call `realpath` on *nix: https://github.com/dotnet/cli/blob/8f7b58dd665f60300221a4eb5910a691749396aa/src/Microsoft.DotNet.MSBuildSdkResolver/MSBuildSdkResolver.cs#L179-L186. I think you'll want to fix that.

src/MSBuildLocator/VisualStudioInstanceQueryOptions.cs
```diff
@@ -13,12 +15,23 @@ public class VisualStudioInstanceQueryOptions
...
 
         /// <summary>
         ///     Discovery types for instances included in the query.
         /// </summary>
         public DiscoveryType DiscoveryTypes { get; set; }
+
+        /// <summary>
+        /// A working directory to use when querying for instances.
+        /// </summary>
+        public string WorkingDir { get; set; } = Environment.CurrentDirectory;
```
DustinCampbell: Should this be `WorkingDirectory` since it's a public API?

src/MSBuildLocator/DiscoveryType.cs
DustinCampbell: Oh, I see. Because it can be passed as an argument to limit the discovery. #Closed

src/MSBuildLocator/DotNetSdkLocationHelper.cs
```diff
@@ -0,0 +1,124 @@
...
+        {
+            Process process = Process.GetCurrentProcess();
```
DustinCampbell: Oh. Maybe I'm wrong about using `realpath` on *nix if this is how we're getting `dotnet`. I can see how this might work for shared framework apps, but how would it work for standalone .NET Core apps?

src/MSBuildLocator/DotNetSdkLocationHelper.cs
jeffkl: We discussed renaming the class in a breaking 2.0 version.  Is that an idea you support?

src/MSBuildLocator/DotNetSdkLocationHelper.cs
jeffkl: I decided to use a `FileInfo` because I'm checking its existence and then using the full path.  If I'm just checking the existence then I use `File.Exists()` generally.

src/MSBuildLocator/DotNetSdkLocationHelper.cs
jeffkl: Yeah it should work as its just a copy/paste.  I don't think standalone apps need MSBuildLocator since they'd need to have the full closure of everything needed in order to use MSBuild.  They'd have to reference a lot more NuGet packages like compilers etc.  I wonder if we can detect that its a standalone app and error out?  Do you know of any way to distinguish at run time?

src/MSBuildLocator/DotNetSdkLocationHelper.cs
jeffkl: This might be a way of detecting standalone app vs shared app.  If the main module isn't `dotnet.exe`, it must be standalone right?  Maybe throw here saying to not use MSBuildLocator?

jeffkl:
I just tested on my Ubuntu machine:

```
jeff@JEFFKL-LINUX16:~/msbuildlocator/samples/BuilderApp$ dotnet exec /home/jeff/msbuildlocator/samples/BuilderApp/bin/Debug/netcoreapp2.0/BuilderApp.dll test.proj 
Sample MSBuild Builder App 1.0.19+g351e77f99f.

1) .NET Core SDK - 2.1.103

Select an instance of MSBuild: 
1
Using VS Instance: .NET Core SDK - 2.1.103


BuildApp running using MSBuild version 15.6.82.30579
/usr/share/dotnet/sdk/2.1.103

Build started.
Project "test.proj" (default targets):
Building with tools version "15.0".
Project file contains ToolsVersion="". This toolset may be unknown or missing, in which case you may be able to resolve this by installing the appropriate version of MSBuild, or the build may have been forced to a particular ToolsVersion for policy reasons. Treating the project as if it had ToolsVersion="15.0". For more information, please see http://go.microsoft.com/fwlink/?LinkId=293424.
Target "Build" in project "/home/jeff/msbuildlocator/samples/BuilderApp/test.proj" (entry point):
Task "Error" skipped, due to false condition; (!$(MSBuildExtensionsPath.StartsWith('$(ProgramFiles(x86))'))) was evaluated as (!True).
Done building target "Build" in project "test.proj".
Done building project "test.proj".
Build succeeded.

Build result: True
```

src/MSBuildLocator/DotNetSdkLocationHelper.cs
DustinCampbell: I'm not sure what the right thing to do would be. @livarcocc, do you have any ideas for what we could do for standalone apps?

src/MSBuildLocator/DotNetSdkLocationHelper.cs
DustinCampbell: No idea. I guess you're right that standalone apps are really out of luck. In that case, I suppose they need to basically include everything needed to run `dotnet build` themselves, including MSBuild and the SDK resolvers.

jeffkl:
Its currently crashing on OS X, I'll look into it later but don't merge yet

DustinCampbell:
FWIW, @livarcocc and I were chatting about this, and we were concerned that it would only work if the app is launched with `dotnet run`.

jeffkl:
I suppose I need to understand how .NET Core apps are launched. 

1. Stand-alone - Build output contains everything needed to run the app.  `dotnet.exe` is renamed to `myapp.exe` so users just run `myapp.exe`
2. Shared app - Build output contains `myapp.dll` and a `myapp.deps.json` with pointers to dependencies.  Users run `dotnet.exe myapp.dll`

Is `dotnet run` different than `dotnet myapp.dll`?  The entry point is still `dotnet.exe` no matter what right?  Can you not P/Invoke into `hostfxr.dll` for a stand-alone app?  Should I just go talk to @livarcocc ?

Thanks!

src/MSBuildLocator/DotNetSdkLocationHelper.cs
```diff
@@ -0,0 +1,124 @@
...
+            }
+
+            return buffer.ToString();
+        }
+
+        private static class Interop
+        {
+            private static readonly bool RunningOnWindows = RuntimeInformation.IsOSPlatform(OSPlatform.Windows);
+
+            public static string realpath(string path)
```
nguerrera: This doesn't appear to be used.

src/MSBuildLocator/DotNetSdkLocationHelper.cs
```diff
@@ -0,0 +1,124 @@
...
+                return RunningOnWindows
+                    ? windows_hostfxr_resolve_sdk(exeDir, workingDir, buffer, bufferSize)
+                    : unix_hostfxr_resolve_sdk(exeDir, workingDir, buffer, bufferSize);
+            }
+
+            [DllImport("hostfxr", EntryPoint = nameof(hostfxr_resolve_sdk), CharSet = CharSet.Unicode, ExactSpelling = true, CallingConvention = CallingConvention.Cdecl)]
+            private static extern int windows_hostfxr_resolve_sdk(string exeDir, string workingDir, [Out] StringBuilder buffer, int bufferSize);
+
+            // CharSet.Ansi is UTF8 on Unix
+            [DllImport("hostfxr", EntryPoint = nameof(hostfxr_resolve_sdk), CharSet = CharSet.Ansi, ExactSpelling = true, CallingConvention = CallingConvention.Cdecl)]
```
nguerrera: Did you try this on Unix yet? I don't believe the P/invoke will just bind to the hostfxr that's already been loaded on non-Windows.

nguerrera:
My main concern is this:  We reserve the right to move things around in the .NET Core SDK directory. It is a long standing problem that everything is in the root with unified deps.json. There is a desire to split msbuild from nuget, etc. I would like this code to document that is depending on this implementation detail and that the mitigation is that this code accepts the risk of getting broken by a layout change. If the layout changes, we must update the locator, not push to revert the layout change. OK?

src/MSBuildLocator/DotNetSdkLocationHelper.cs
nguerrera: I see above that you had it working on Ubuntu. I am confused now. Maybe it was only on OS X where I saw this behavior, See https://github.com/dotnet/core-setup/pull/4301/commits/d438d9838afcb7a15e86348ff1a2f8084e54e206 where I had to copy libhostfxr for tests.

nguerrera: Thinking more, I'm 99% sure I've seen this behavior difference on Linux too. I'm very confused.

src/MSBuildLocator/DotNetSdkLocationHelper.cs
nguerrera: I wouldn't expect standalone apps to fail here, but instead fail to find any sdks below.

src/MSBuildLocator/DotNetSdkLocationHelper.cs
nguerrera: I can repro it working on Ubuntu and failing on Mac OS X. From the discussion [here](https://github.com/dotnet/coreclr/issues/12707), I would not expect it to work on Ubuntu either, but it does.

jeffschwMSFT:
cc @swaroop-sridhar

jeffkl:
@nguerrera I never figured out why it crashes on OS X, do you have any idea why?

nguerrera:
I have an idea, but everything I read suggested it would apply equally to Mac OS X and Linux so my idea may be wrong.

nguerrera:
The failure on Mac OS X matches exactly what I expected the failure to be on Linux, but it works on Linux.

nguerrera:
In a related issue around P/Invoke to existing dlls, I saw the behavior we're seeing here on Mac on RHEL6 as well. So in addition to not working on Mac, it does not work on all Linuxes.

jeffschwMSFT:
cc @vitek-karas @AaronRobinsonMSFT

src/MSBuildLocator/DotNetSdkLocationHelper.cs
AaronRobinsonMSFT: If my opinion matters, I would very much support changing the class name.

src/MSBuildLocator/DotNetSdkLocationHelper.cs
AaronRobinsonMSFT: I don't see how `process` can ever be `null`. I don't think that API is capable of returning null so why the check below? That then begs the question can `MainModule` be null, which I don't think that is possible either since I think it will throw on access if it can't be found. Are there cases where `null` is possible?

As a general aside this is my biggest gripe with the `?` operator we now have potential unnecessary `null` checks everywhere due to copy/pasting :-/

src/MSBuildLocator/DotNetSdkLocationHelper.cs
```diff
@@ -0,0 +1,124 @@
...
+            string dotNetExePath = process?.MainModule?.FileName;
+
+            if (String.IsNullOrWhiteSpace(dotNetExePath))
+            {
+                return null;
+            }
+
+            string dotNetExeDirectory = Path.GetDirectoryName(dotNetExePath);
+
+            if (String.IsNullOrWhiteSpace(dotNetExeDirectory) || !Directory.Exists(dotNetExeDirectory))
```
AaronRobinsonMSFT: The `String.IsNullOrWhiteSpace()` check isn't needed since the `Directory.Exists()` API handles both cases.

src/MSBuildLocator/DotNetSdkLocationHelper.cs
```diff
@@ -0,0 +1,124 @@
...
+            {
+                // hostfxr string encoding is platform -specific so dispatch to the
+                // appropriately annotated P/Invoke for the current platform.
+                return RunningOnWindows
+                    ? windows_hostfxr_resolve_sdk(exeDir, workingDir, buffer, bufferSize)
+                    : unix_hostfxr_resolve_sdk(exeDir, workingDir, buffer, bufferSize);
+            }
+
+            [DllImport("hostfxr", EntryPoint = nameof(hostfxr_resolve_sdk), CharSet = CharSet.Unicode, ExactSpelling = true, CallingConvention = CallingConvention.Cdecl)]
+            private static extern int windows_hostfxr_resolve_sdk(string exeDir, string workingDir, [Out] StringBuilder buffer, int bufferSize);
```
AaronRobinsonMSFT: I have no real dog in this fight, but using `StringBuilder` as a variable output buffer has terrible performance. If this is a concern in this code path we should fix the native API, but I believe that you could probably get away with a `ref string` and have better performance with no native API change.

PR #34, opened 2018-05-16
Fix potential null-reference if "VSCMD_VER" is not set

DustinCampbell:
In cases where "VSINSTALLDIR" is set but "VSCMD_VER" is not, getting the dev console instance could null-ref.


DustinCampbell:
FYI that I don't have access to merge this if you guys want to take it.

PR #35, opened 2018-05-19
Register from path

cdmihai:
Useful if you know exactly where MSBuild is, for example if you want to point it to the msbuild repo bootstrap directory.

### DIFF
--- a/samples/BuilderApp/Program.cs
+++ b/samples/BuilderApp/Program.cs
@@ -30,14 +30,14 @@ namespace BuilderApp
             // safely call code that use MSBuild types (in the Builder class).
             if (msbuildDeploymentToUse.VSInstance != null)
             {
-                Console.WriteLine($"Using MSBuild deployment from VS Instance: {msbuildDeploymentToUse.VSInstance.Name} - {msbuildDeploymentToUse.VSInstance.Version}");
+                Console.WriteLine($"Using MSBuild from VS Instance: {msbuildDeploymentToUse.VSInstance.Name} - {msbuildDeploymentToUse.VSInstance.Version}");
                 Console.WriteLine();
 
                 MSBuildLocator.RegisterInstance(msbuildDeploymentToUse.VSInstance);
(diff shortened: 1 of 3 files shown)


src/MSBuildLocator/MSBuildLocator.cs
```diff
@@ -69,6 +69,16 @@ public static void RegisterInstance(VisualStudioInstance instance)
...
 
+            RegisterMSbuildPath(instance.MSBuildPath);
+        }
+
+        /// <summary>
+        ///     Add assembly resolution for Microsoft.Build core dlls in the current AppDomain from the specified
+        ///     path.
+        /// </summary>
+        /// <param name="msbuildPath"></param>
+        public static void RegisterMSbuildPath(string msbuildPath)
```
DustinCampbell: Should the casing be `RegisterMSBuildPath` with a capital **B**? That matches other APIs here.

samples/BuilderApp/Program.cs
```diff
@@ -24,14 +24,24 @@ private static void Main(string[] args)
...
-            var instanceToUse = AskWhichVisualStudioInstanceToUse(instances);
+            var msbuildDeploymentToUse = AskWhichMSBuildToUse(instances);
 
-            // Calling RegisterInstance will subscribe to AssemblyResolve event. After this we can now
+            // Calling Register methods will subscribe to AssemblyResolve event. After this we can
             // safely call code that use MSBuild types (in the Builder class).
-            MSBuildLocator.RegisterInstance(instanceToUse);
+            if (msbuildDeploymentToUse.VSInstance != null)
+            {
+                Console.WriteLine($"Using MSBuild deployment from VS Instance: {msbuildDeploymentToUse.VSInstance.Name} - {msbuildDeploymentToUse.VSInstance.Version}");
```
rainersigwald: I would not use the word "deployment" here. Just say `Using MSBuild from...`

Issue #42, opened 2018-08-16
RegisterDefault() throws ArgumentNullException if VS not installed


yaakov-h:
```
System.ArgumentNullException : Value cannot be null.
Parameter name: instance
```

This should probably be an `InvalidOperationException` or similar.

PR #43, opened 2018-09-02
Better error handling

dfederm:
Closes: #42

### DIFF
--- a/src/MSBuildLocator/MSBuildLocator.cs
+++ b/src/MSBuildLocator/MSBuildLocator.cs
@@ -76,6 +76,15 @@ namespace Microsoft.Build.Locator
         public static VisualStudioInstance RegisterDefaults()
         {
             var instance = GetInstances().FirstOrDefault();
+            if (instance == null)
+            {
+                var error = "No instances of MSBuild could be detected." +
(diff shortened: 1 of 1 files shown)


src/MSBuildLocator/MSBuildLocator.cs
```diff
@@ -85,6 +96,11 @@ public static void RegisterInstance(VisualStudioInstance instance)
         /// </param>
         public static void RegisterMSBuildPath(string msbuildPath)
         {
+            if (string.IsNullOrWhiteSpace(msbuildPath))
```
cdmihai: Maybe check for existence too?

cdmihai: I see that there's a list of msbuild dll names. Checking that those dlls actually exist in `msbuildPath` might help debugging when they don't exist.

dfederm: I'm adding the existence check. For checking each dll though, I'm worried there are some cases where potentially not all dlls are required by the caller, so if they're missing then it's fine. The current assembly resolver just ignores missing dlls.

Issue #5, opened 2017-09-14
There should be an MSBuildLocator.Dispose() method


DustinCampbell:
Currently, MSBuildLocator hooks the AppDomain.AssemblyResolve event and that hook will stay in place for the lifetime of the AppDomain. However, in some scenarios, it is necessary to unhook and this event if several MSBuildLocators might be constructed (for example, in a unit testing scenario).

AndyGerlicher:
Makes sense. I thought about that but didn't get around to it (methods are all static so not quite that easy). One issue though is by the time anything is loaded it's really too late to remove that hook in a meaningful way. And it won't get called for those assemblies on that AppDomain anyway if they're already loaded.

tibel:
For unit tests a property `IsRegistered` would help.
Subsequent calls to `RegisterInstance()` throw `InvalidOperationException`.

PR #44, opened 2018-09-02
Add automatic and manual unregistration

dfederm:
Closes: #5


src/MSBuildLocator/MSBuildLocator.cs
```diff
@@ -85,10 +108,9 @@ public static void RegisterInstance(VisualStudioInstance instance)
         /// </param>
         public static void RegisterMSBuildPath(string msbuildPath)
         {
-            var loadedMSBuildAssemblies = AppDomain.CurrentDomain.GetAssemblies().Where(IsMSBuildAssembly);
-            if (loadedMSBuildAssemblies.Any())
+            if (!CanRegister)
```
dfederm: Interestingly, this fixes a minor bug where you can register multiple time so long as no actual assemblies have been loaded into the app domain yet.

cdmihai:
Can you write a few unit tests for these? :)

cdmihai:
OK, there's no unit tests whatsoever. Feel free to ignore :)

dfederm:
Yea, unit testing assembly resolution is a little difficult :(. We'd need to stub out both the AssemblyResolve event and the Assembly.Load calls I guess, but at some point that just adds too much complexity for such a simple implementation...

PR #45, opened 2018-09-09
Upgrade dependencies

dfederm:


samples/BuilderApp/BuilderApp.csproj
```diff
@@ -12,8 +12,8 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.Build" Version="15.5.180" />
-    <PackageReference Include="Microsoft.Build.Framework" Version="15.5.180" />
+    <PackageReference Include="Microsoft.Build" Version="15.8.166" />
+    <PackageReference Include="Microsoft.Build.Framework" Version="15.8.166" />
```
rainersigwald: I think we should deliberately leave this low, probably with a comment. Part of the point of this package is that you don't have to stay on the MSBuild-update treadmill.

dfederm: Just greenkeeping. Sure, I'll back out the MSBuild upgrade and add a comment. This is just the sample app, but I get the idea.

PR #51, opened 2018-10-12
Support locating .NET Core SDK instances using `dotnet --info`

JoeRobich:
This PR is an adaptation of the work @jeffkl submitted in https://github.com/Microsoft/MSBuildLocator/pull/33. However, the method of determining the the dotnet base path is different.

This PR uses `dotnet --info` to get the proper base path in a platform agnostic manner. This method is used by Omnisharp-roslyn (https://github.com/OmniSharp/omnisharp-roslyn/blob/master/src/OmniSharp.MSBuild/SdksPathResolver.cs) and docfx (https://github.com/dotnet/docfx/blob/dev/src/Microsoft.DocAsCode.Metadata.ManagedReference/MsBuildEnvironmentScope.cs).


src/MSBuildLocator/Microsoft.Build.Locator.csproj
```diff
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <OutputType>Library</OutputType>
-    <TargetFramework>net46</TargetFramework>
+    <TargetFrameworks>net46;netstandard2.0</TargetFrameworks>
```
DustinCampbell: Should this be `netcoreapp2.0` rather than `netstandard2.0`. It seems weird that a `netstandard2.0` version of this package would only locate MSBuild via .NET Core SDK -- especially since `netstandard2.0` is supposed to be compatible with `net461`. Making the package `netstandard2.0` signals support that it probably shouldn't.

DustinCampbell:
Invoking @rainersigwald and @AndyGerlicher here. Also, @nguerrera.

src/MSBuildLocator/DotNetSdkLocationHelper.cs
```diff
@@ -0,0 +1,102 @@
...
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.IO;
+using System.Text.RegularExpressions;
+
+namespace Microsoft.Build.Locator
+{
+    internal static class DotNetSdkLocationHelper
+    {
+        private static readonly Regex DotNetBasePathRegex = new Regex("Base Path:(.*)$", RegexOptions.Compiled | RegexOptions.Multiline);
```
nguerrera: Is it worth compiling this regex? My intuition is that the process creation will dwarf the cost.

This whole thing is generally expected to be called once per caller process / app domain, right?

src/MSBuildLocator/MSBuildLocator.cs
```diff
@@ -212,6 +238,77 @@ public static void Unregister()
...
+                    "Loaded MSBuild assemblies: " +
+                    loadedAssemblyList;
+
+                throw new InvalidOperationException(error);
+            }
+
+            foreach (var msBuildAssembly in s_msBuildAssemblies)
+            {
+                var targetAssembly = Path.Combine(msbuildPath, msBuildAssembly + ".dll");
+                Assembly.LoadFrom(targetAssembly);
```
nguerrera: I'm not sure about how LoadFrom works on Core, but I vaguely recall that it might require the assembly to be present in deps.json. If so, that might still be mostly ok because the use case is to bind static compilation against msbuild to a dynamic location. 

But I'm unsure how this would behave if you had an indirection from a main app that knew nothing of msbuild to a plug-in that did. It seems you would want to resolve msbuild into the caller's assemblyloadcontext. I don't know if this accomplishes that or not.

@jeffschwMSFT and/or team should review this.

nguerrera:
Oh, and as I've stated in all previous PRs on the subject: the location of MSBuild dlls within the SDK base path is subject to change, so the library here should be prepared to be updated on that. The base path is the location of dotnet.dll and we actually have a desire to make that different from the location of msbuild.dll.

I'm now thinking we could come up with something more resilient by adding something contractual to the SDK. For example, we could have `MSBuild Base Path:` in info and the locator could prefer that over `Base Path` if found. I've filed https://github.com/dotnet/cli/issues/10195

You don't need to block on that, but we should discuss it now while this is fresh in our minds.

JoeRobich:
Thanks to @nguerrera. I removed the assembly pre-load for the .net core assemblies. Listening for `AssemblyLoadContext.Default.Resolving` is the proper way to handle the assembly load on .Net Core.

nguerrera:
I didn't actually expect that to fix the issue we were debugging with mismatched nuget versions so I'll have to check why it seems so to have done so 

Also I think you want to resolve into the load context of the caller and not necessarily default. I'm not sure that can be done without an api change to locator.

DustinCampbell:
> Listening for AssemblyLoadContext.Default.Resolving is the proper way to handle the assembly load on .Net Core.

FWIW, `AppDomain.CurrentDomain.AssemblyResolve` has worked since netcoreapp2.0. IIRC, that particular event is shimmed to use `AssemblyLoadContext.Default`. So, you should be to just use that event and remove #ifdefs.

DustinCampbell:
Invoking @AndyGerlicher and @rainersigwald again.

JoeRobich:
> FWIW, AppDomain.CurrentDomain.AssemblyResolve has worked since netcoreapp2.0. IIRC, that particular event is shimmed to use AssemblyLoadContext.Default. So, you should be to just use that event and remove #ifdefs.

Hmm, I wonder if there is an issue in the shim then, because that is how I initially coded things. However, What I ended up with was a `Process is terminating due to StackOverflowException`. This issue had previously been logged [here](https://github.com/Microsoft/msbuild/issues/3352). The version that uses `AssemblyLoadContext` does not throw an exception. =/

jeffschwMSFT:
@JoeRobich can you file a bug on dotnet/coreclr for any of the issues that you feel are related to AssemblyLoadContext, we would be happy to take a look.  If the repro's can be extracted from msbuild, that would be best.  If they are only reproable with msbuild, perhaps they can remain in this repo.  Can you extract the use of AssemblyResolveEvent/AssemblyLoadContext so that I can take a look and see if there is a reasonable explanation for the SO?

DustinCampbell:
@jeffschwMSFT: I think there's a repro project in https://github.com/Microsoft/msbuild/issues/3352.

jeffschwMSFT:
@DustinCampbell we are chatting offline about the SO issue.  The underlying issue is that msbuild's assembly resolve event uses LoadAssemblyByName which introduces a cycle.

JoeRobich:
@DustinCampbell I've submitted a fix for the SO that I was seeing with `AssemblyResolve` (https://github.com/Microsoft/msbuild/pull/3900), but I think it would be a while before we could rely on it for .NET Core.

DustinCampbell:
Pinging this PR. What are the next steps for merging this?

nguerrera:
@DustinCampbell There's an issue with resolution of task dependencies. I added you to a thread. @JoeRobich is there anything else?

JoeRobich:
The only issue we've run across is the dependencies issue. For us the short term fix was to add Newtonsoft.Json as a dependency for dotnet-format.

Without it MSBuilds CoreCLRAssemblyLoader will look for them, but with a strict version match check. This fails in the 3.0.0 sdk for us because the version shipping with the sdk is higher than the version the task was built against.

nguerrera:
Confused how NewtonSoft.Json alone fixed things when the thread was about loading nuget dlls depended upon by sdk build tasks. Is it just that dotnet format doesn't actually kick off a build or use of these sdk tasks?

nguerrera:
I'll review it again now.

JoeRobich:
@nguerrera We don't actually perform a build. I was just using it so I could open projects and solutions in a roslyn MSBuildWorkspace.

src/MSBuildLocator/MSBuildLocator.cs
```diff
@@ -209,7 +261,83 @@ public static void Unregister()
...
+        ///     Load Microsoft.Build core dlls into the current AppDomain from the specified path.
+        /// </summary>
+        /// <param name="msbuildPath">
+        ///     Path to the directory containing a deployment of MSBuild binaries.
+        ///     A minimal MSBuild deployment would be the publish result of the Microsoft.Build.Runtime package.
+        ///
+        ///     In order to restore and build real projects, one needs a deployment that contains the rest of the toolchain (nuget, compilers, etc.).
+        ///     Such deployments can be found in installations such as Visual Studio or dotnet CLI.
+        /// </param>
+        public static void LoadFromMSBuildPath(string msbuildPath)
```
nguerrera: >LoadFromMSBuildPath [](start = 27, length = 19)

Why is this method still needed?

src/MSBuildLocator/MSBuildLocator.cs
```diff
@@ -23,7 +27,11 @@ public static class MSBuildLocator
             "Microsoft.Build.Utilities.Core"
         };
 
+#if NET46
         private static ResolveEventHandler registeredHandler;
+#else
+        private static Func<AssemblyLoadContext, AssemblyName, Assembly> registeredHandler;
```
nguerrera: >registeredHandler [](start = 73, length = 17)

nit: Pre-existing, but there's an inconsistency with resect to s_ prefix for static variables.

src/MSBuildLocator/MSBuildLocator.cs
```diff
@@ -209,7 +261,83 @@ public static void Unregister()
...
+                    loadedAssemblyList;
+
+                throw new InvalidOperationException(error);
+            }
+
+            foreach (var msBuildAssembly in s_msBuildAssemblies)
+            {
+                var targetAssembly = Path.Combine(msbuildPath, msBuildAssembly + ".dll");
+                var assemblyName = AssemblyName.GetAssemblyName(targetAssembly);
+                Assembly.Load(assemblyName);
```
nguerrera: >Assembly.Load(assemblyName); [](start = 16, length = 28)

This doesn't do what the method documentation says.

src/MSBuildLocator/MSBuildLocator.cs
```diff
@@ -158,29 +185,54 @@ public static void RegisterMSBuildPath(string msbuildPath)
...
+                {
+                    if (loadedAssemblies.TryGetValue(assemblyName.FullName, out var assembly))
                     {
-                        var targetAssembly = Path.Combine(msbuildPath, assemblyName.Name + ".dll");
-                        if (File.Exists(targetAssembly))
+                        return assembly;
+                    }         
+            
+                    return TryLoadAssembly(assemblyName);
+                }
```
nguerrera: Consider extracting a method for the #if/#else cases. Or could the commonality be moved to TryLoadAssembly?

src/MSBuildLocator/MSBuildLocator.cs
JoeRobich: It isn't. Removing.

src/MSBuildLocator/MSBuildLocator.cs
```diff
@@ -103,6 +119,16 @@ public static void RegisterInstance(VisualStudioInstance instance)
                 throw new ArgumentNullException(nameof(instance));
             }
 
+            if (instance.DiscoveryType == DiscoveryType.DotNetSdk)
+            {
+                // Since there can be multiple DotNet SDKs on a machine
+                // the MSBuild environment variable are likely not set.
+                // To work around this problem we configure the necessary
+                // variables for a DotNet SDK deployment of MSBuild.
+                // Workaround for https://github.com/dotnet/docfx/issues/1752
```
nguerrera: This is not really working around a bug. Certainly not a docfx one.

I think it would be simpler to explain that normally the dotnet cli sets up these environment variables with msbuild is invoked via `dotnet`, but we are using msbuild dlls directly and therefore need to mimic that.

src/MSBuildLocator/MSBuildLocator.cs
```diff
@@ -209,7 +261,83 @@ public static void Unregister()
...
+        {
+            const string MSBUILD_EXE_PATH = nameof(MSBUILD_EXE_PATH);
+            const string MSBuildExtensionsPath = nameof(MSBuildExtensionsPath);
+            const string MSBuildSDKsPath = nameof(MSBuildSDKsPath);
+
+            var variables = new Dictionary<string, string>
+            {
+                [MSBUILD_EXE_PATH] = dotNetSdkPath + "MSBuild.dll",
+                [MSBuildExtensionsPath] = dotNetSdkPath,
+                [MSBuildSDKsPath] = dotNetSdkPath + "Sdks"
```
nguerrera: Some tasks will also expect DOTNET_HOST_PATH to be set on Core to the path to dotnet[.exe]

nguerrera: This is not needed since you rely on dotnet being on PATH.

src/MSBuildLocator/DotNetSdkLocationHelper.cs
```diff
@@ -0,0 +1,105 @@
...
+            {
+                return null;
+            }
+
+            if (!File.Exists(Path.Combine(dotNetSdkPath, "Microsoft.Build.dll")))
+            {
+                return null;
+            }
+
+            FileInfo dotNetAssemblyPath = new FileInfo(Path.Combine(dotNetSdkPath, "dotnet.dll"));
```
rainersigwald: It's a little unfortunate to take the `dotnet.dll` version here, since that's tied to the .net core _runtime_ version, where the interesting thing here is the _sdk_ version.

rainersigwald: I was hoping that the SDK version would be easy to extract from `dotnet --info` but it's not, really:

```
S:\Locator>dotnet --info
.NET Core SDK (reflecting any global.json):
 Version:   3.0.100-preview-009812
 Commit:    e3abf6e935

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.17763
 OS Platform: Windows
 RID:         win10-x64
 Base Path:   C:\Program Files\dotnet\sdk\3.0.100-preview-009812\
```

I don't want to assume that there would never be anything between the `.NET Core SDK` line and the `Version:` line, and `Version:` isn't unique  (there's a host version too). So I guess this is ok for now.

src/MSBuildLocator/VisualStudioInstanceQueryOptions.cs
```diff
@@ -13,12 +15,23 @@ public class VisualStudioInstanceQueryOptions
...
+#endif
         };
 
         /// <summary>
         ///     Discovery types for instances included in the query.
         /// </summary>
         public DiscoveryType DiscoveryTypes { get; set; }
+
+        /// <summary>
+        ///     A working directory to use when querying for instances.
```
rainersigwald: Yeah, this one is pretty awkward: because `global.json` can cause different SDKs to load, you may have to tear down your process to build multiple projects. The only way around that I can think of is having a different AssemblyLoadContext for each, and I don't understand the interop problems that might result. I think we need to make sure we're clear on this one.

```suggestion
        ///     Working directory to use when querying for instances. Ensure it is the project directory to pick up the right global.json.
```

src/MSBuildLocator/DotNetSdkLocationHelper.cs
nguerrera: What is this version used for? The version of dotnet.dll is not related to the runtime, it is the version of dotnet/cli repo (for 2.x this is the same as the sdk version, for 3.x it is not.)

nguerrera: The SDK version is relatively easy to extract from the .version file in the Base Path.

nguerrera: Also, for `dotnet --info` to locate an SDK, the unqualified directory name will be its version.

nguerrera: Not sure which implementation detail is better to depend upon. There is also the fact that the API here wants a System.Version and not a semantic version.

JoeRobich: @rainersigwald So this code is looking at the dotnet.dll within the SDK directory. Opening file properties on a few of these locally shows the FileVersion matches the dotnet sdk version. 
image

rainersigwald: Yeah, this is more and more confusing: 
image

Maybe we could just feed the name of the folder into the VS instance name? I also lamented that we have a `System.Version` rather than something useful in that structure.

nguerrera: Right, as I said "for 2.x this is the same as the sdk version, for 3.x it is not". In 3.x dotnet.dll is built in dotnet/cli repo and inserted into dotnet/core-sdk as a binary. The version number of dotnet/cli is not a customer facing number.

nguerrera: (Complete aside: that light properties dialog over a dark explorer is bothering my eyes.)

JoeRobich: @nguerrera Took your suggestion to use the .version file. Thanks

src/MSBuildLocator/DotNetSdkLocationHelper.cs
```diff
@@ -0,0 +1,119 @@
...
+using System.Diagnostics;
+using System.IO;
+using System.Text.RegularExpressions;
+
+namespace Microsoft.Build.Locator
+{
+    internal static class DotNetSdkLocationHelper
+    {
+        private static readonly Regex DotNetBasePathRegex = new Regex("Base Path:(.*)$", RegexOptions.Multiline);
+        private static readonly Regex VersionRegex = new Regex(@"^(\d+).(\d+).(\d+)", RegexOptions.Multiline);
```
rainersigwald:
```suggestion
        private static readonly Regex VersionRegex = new Regex(@"^(\d+)\.(\d+)\.(\d+)", RegexOptions.Multiline);
```

I lucked into an extremely confusing error mode! The SDK I had installed happened to fit the earlier pattern, since `.` could match the non-decimal hex digits:

```
91e160c7f04c8f966895d441714b6dd6697a2f94
2.1.504
win-x64
```

gave me version `91.160.7`.

JoeRobich: haha thanks. It's been a while since I regex'd.

nguerrera:
:clap:

rainersigwald:
A package with this in it is available at https://dotnet.myget.org/feed/msbuild/package/nuget/Microsoft.Build.Locator/1.2.2. If we can get some upvotes that it works, we'll push it to nuget.org.

JoeRobich:
@rainersigwald I verified with a build of dotnet-format running on Win10 and Ubuntu

rainersigwald:
1.2.2 is now on nuget.org: https://www.nuget.org/packages/Microsoft.Build.Locator/1.2.2

PR #74, opened 2019-08-12
Set up official build YAML

rainersigwald:


azure-pipelines.yml
```diff
@@ -0,0 +1,74 @@
+pool:
+  name: VSEng-MicroBuildVS2017
+  demands: msbuild
```
Forgind, line 3: Does this have to be a boolean?

azure-pipelines.yml
```diff
@@ -0,0 +1,74 @@
...
+    platform: '$(BuildPlatform)'
+    configuration: '$(BuildConfiguration)'
+    msbuildArguments: '/t:Pack /restore -binaryLogger:$(Build.SourcesDirectory)\Pack.binlog'
+
+- task: CopyFiles@2
+  displayName: 'Collect Symbols'
+  inputs:
+    SourceFolder: src/MSBuildLocator
+    Contents: '**\*.pdb'
+    TargetFolder: '$(Build.ArtifactStagingDirectory)/symbols'
```
Forgind, line 29: slashes seem to be going both ways. This is windows, right?

Forgind (commented):
https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=2936208

It looks like the build was not found? Why is that?

azure-pipelines.yml
rainersigwald: Not sure I understand your question. This means "the thing this runs on must have the capability named `msbuild`", which is probably a bit redundant since everything in the pool does but it was part of the UI->YAML conversion.

azure-pipelines.yml
rainersigwald: Yeah, our official build is on Windows, but I don't think this matters.

azure-pipelines.yml
```diff
@@ -0,0 +1,81 @@
...
+    ArtifactType: Container
+
+- task: CopyPublishBuildArtifacts@1
+  displayName: 'Upload MicroBuildOutputs to drop'
+  inputs:
+    CopyRoot: '$(Build.StagingDirectory)\MicroBuild\Output'
+    Contents: '**'
+    ArtifactName: MicroBuildOutputs
+    ArtifactType: Container
+  continueOnError: true
```
Forgind, line 70:
```suggestion
- task: PublishPipelineArtifact@1
  displayName: 'Upload MicroBuildOutputs to drop'
  inputs:
    targetPath: '$(Build.StagingDirectory)\MicroBuild\Output'
    artifactName: MicroBuildOutputs
  continueOnError: true
```

azure-pipelines.yml
```diff
@@ -29,45 +29,43 @@ steps:
...
     ArtifactName: MicroBuildOutputs
-    ArtifactType: Container
   continueOnError: true
 
-- task: CopyPublishBuildArtifacts@1
+- task: PublishBuildArtifacts@1
   displayName: 'Upload logs to drop'
   inputs:
-    Contents: '**\*log'
+    PathToPublish: '$(Build.StagingDirectory)\src\MSBuildLocator\bin'
```
rainersigwald, line 67: This doesn't look right to me?

Forgind: Copy-paste error 😅

That said, I'm not sure how to convert that one to the new style. The build with the penultimate commit worked, so we can go back to that if you want. I just wanted to also not use deprecated things if possible even though they weren't failing.

Forgind:
@rainersigwald,
Ok, so I think it's finally in better shape. It seems to have succeeded. I took out the other instances of the deprecated task. Does it look good to you?

https://dev.azure.com/devdiv/DevDiv/_build/results?buildId=6479413&view=logs&j=12f1170f-54f2-53f3-20dd-22fc7dff55f9&t=57a0d086-873a-57cb-8745-4852304b19bb

And then the one other thing I think we're missing is getting M2 approval to fix the signing thing. Is there anything else?

azure-pipelines.yml
```diff
@@ -0,0 +1,91 @@
...
+  inputs:
+    PathToPublish: '$(Build.ArtifactStagingDirectory)\pkg'
+    ArtifactName: pkg
+
+- task: PublishPipelineArtifact@1
+  displayName: 'Upload MicroBuildOutputs to drop'
+  inputs:
+    TargetPath: '$(Build.StagingDirectory)\bin'
+    ArtifactName: MicroBuildOutputs
+  continueOnError: true
```
rainersigwald, line 75: The `MicroBuildOutputs` of your test build don't look like the `MicroBuildOutputs` of the existing builds, which I think is masking that something is wrong with the "Enable Source Server" step that should be zipping up the sources.

Forgind (approved):

PR #76, opened 2019-08-28
Filter VS instances to those containing MSBuild

### LINKED ISSUE #73: RegisterDefaults should not return a Visual Studio instance without MSBuild
For calculating code metrics I have included Microsoft.CodeAnalysis.Metrics 2.9.4 nuget package to my csproj project.
Before executing msbuild from the command line, I have called  "nuget.exe restore", and it correctly found MSBuild and restore package:

> "MSBuild auto-detection: using msbuild version '16.200.19.32702' from 'C:\Program Files (x86)\Microsoft Visual Studio\2019\BuildTools\MSBuild\Current\bin'.
> Restoring NuGet package Microsoft.CodeAnalysis.Metrics.2.9.4. ..."

Then, from command line I am calling msbuild with /t:Metrics option:

> "C:\Program Files (x86)\Microsoft Visual Studio\2019\BuildTools\MSBuild\Current\bin\MSBuild.exe" /p:Configuration=Debug /t:Metrics My.sln "

which ended with following error:

> "Unhandled Exception: System.ArgumentException: Directory "C:\Program Files (x86)\Microsoft Visual Studio\2019\TestAgent\MSBuild\Current\Bin" does not exist
>   Parameter name: msbuildPath
>      at Microsoft.Build.Locator.MSBuildLocator.RegisterMSBuildPath(String msbuildPath)
>      at Microsoft.Build.Locator.MSBuildLocator.RegisterDefaults()
>      at Metrics.Program.<RunAsync>d__1.MoveNext()
>   --- End of stack trace from previous location where exception was thrown ---
>      at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
>      at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)"

It looks like  Metrics.exe looking for msbuild in wrong location, but nugget.exe in correct one?

rainersigwald:
Avoid returning Visual Studio instances that don't have MSBuild
installed, for instance SSDT or Test Agent.

I'm basing the check on the internal component that pulls in the GitHub
MSBuild package because it also ensures the presence of Roslyn.

This should fix #73, but I haven't tested in that scenario (Test Agent installed) yet.

I don't _think_ this is a breaking change warranting a semver major version bump. I think "the thing called MSBuildLocator returned instances of VS that didn't include MSBuild" is just a bug.

### DIFF
--- a/src/MSBuildLocator/VisualStudioLocationHelper.cs
+++ b/src/MSBuildLocator/VisualStudioLocationHelper.cs
@@ -43,8 +43,8 @@ namespace Microsoft.Build.Locator
                     e.Next(1, instances, out fetched);
                     if (fetched <= 0) continue;
 
-                    var instance = instances[0];
-                    var state = ((ISetupInstance2) instance).GetState();
+                    var instance = (ISetupInstance2) instances[0];
+                    InstanceState state = instance.GetState();
 
                     if (!Version.TryParse(instance.GetInstallationVersion(), out Version version))
                         continue;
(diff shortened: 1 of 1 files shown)


src/MSBuildLocator/VisualStudioLocationHelper.cs
```diff
@@ -43,21 +43,36 @@ internal static IList<VisualStudioInstance> GetInstances()
                     e.Next(1, instances, out fetched);
                     if (fetched <= 0) continue;
 
-                    var instance = instances[0];
-                    var state = ((ISetupInstance2) instance).GetState();
+                    var instance = (ISetupInstance2) instances[0];
+                    var state = instance.GetState();
```
Forgind: If you're making little changes like this, may as well change "var" to ISetupInstance2

Forgind: nit: change var to proper type

PR #79, opened 2019-11-29
Return all installed DotNetSDK versions

kurtcodemander:
Returns a list of all installed dotnet SDKs as reported by dotnet --info instead of only the latest

### DIFF
--- a/samples/BuilderApp/BuilderApp.csproj
+++ b/samples/BuilderApp/BuilderApp.csproj
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFrameworks>net472;netcoreapp2.1</TargetFrameworks>
+    <TargetFrameworks>net471;netcoreapp2.1</TargetFrameworks>
     <CopyLocalLockFileAssemblies>false</CopyLocalLockFileAssemblies>
     <IsPackable>false</IsPackable>
   </PropertyGroup>
--- a/samples/BuilderApp/BuilderApp.csproj
+++ b/samples/BuilderApp/BuilderApp.csproj
@@ -24,4 +24,8 @@
     <ProjectReference Include="..\..\src\MSBuildLocator\Microsoft.Build.Locator.csproj" />
   </ItemGroup>
 
+  <!-- Explicitly import the file that gives auto binding redirects for package users.
+       Not necessary if you use the package! -->
+  <Import Project="..\..\src\MSBuildLocator\build\Microsoft.Build.Locator.props"/>
+
 </Project>
--- a/samples/BuilderApp/BuilderApp.csproj
+++ b/samples/BuilderApp/BuilderApp.csproj
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFrameworks>net471;netcoreapp2.1</TargetFrameworks>
+    <TargetFrameworks>net472;netcoreapp2.1</TargetFrameworks>
     <CopyLocalLockFileAssemblies>false</CopyLocalLockFileAssemblies>
     <IsPackable>false</IsPackable>
   </PropertyGroup>
--- a/src/MSBuildLocator/DotNetSdkLocationHelper.cs
+++ b/src/MSBuildLocator/DotNetSdkLocationHelper.cs
@@ -117,25 +117,29 @@ namespace Microsoft.Build.Locator
 
             var outputString = string.Join(Environment.NewLine, lines);
 
-            var lineSdkIndex = lines.FindIndex(line => line.Contains(".NET Core SDKs installed"));
+            var matched = DotNetBasePathRegex.Match(outputString);
+            if (!matched.Success)
+            {
+                return null;
+            }
+
+            var lineSdkIndex = lines.FindIndex(line => line.Contains(".NET Core SDKs installed") || line.Contains(".NET SDKs installed"));
 
             if (lineSdkIndex != -1)
             {
                 lineSdkIndex++; 
 
-                while (lineSdkIndex < lines.Count && !lines[lineSdkIndex].Contains(".NET Core runtimes installed"))
+                while (lineSdkIndex < lines.Count && (!lines[lineSdkIndex].Contains(".NET Core runtimes installed") && !lines[lineSdkIndex].Contains(".NET runtimes installed")))
                 {
-                    var ma = SdkRegex.Match(lines[lineSdkIndex]);
+                    var sdkMatch = SdkRegex.Match(lines[lineSdkIndex]);
 
-                    if (!ma.Success)
+                    if (!sdkMatch.Success)
                         break;
 
-                    var version = ma.Groups[1].Value.Trim();                     
-                    var path = ma.Groups[2].Value.Trim();
+                    var version = sdkMatch.Groups[1].Value.Trim();                     
+                    var path = sdkMatch.Groups[2].Value.Trim();
 
-                    path = Path.Combine(path, version) + "\\";
-
-                    //basePaths.Add(path);
+                    path = Path.Combine(path, version) + "\\";                    
 
                     // We insert at index 0 so that instance list will be sorted descending so that instances.FirstOrDefault() 
                     // will always return the latest installed version of dotnet SDK 
@@ -143,17 +147,8 @@ namespace Microsoft.Build.Locator
 
                     lineSdkIndex++;
                 }
-            }
-
-            var matched = DotNetBasePathRegex.Match(outputString);
-            if (!matched.Success)
-            {
-                return null;
-            }
-
-            //basePaths.Add(matched.Groups[1].Value.Trim());
-            basePaths.Insert(0, matched.Groups[1].Value.Trim());    // Insert at index 0 to not break FirstOrDefault() returning the latest SDK
-
+            }            
+            
             return basePaths;
         }
     }
--- a/src/MSBuildLocator/DotNetSdkLocationHelper.cs
+++ b/src/MSBuildLocator/DotNetSdkLocationHelper.cs
@@ -15,7 +15,7 @@ namespace Microsoft.Build.Locator
     {
         private static readonly Regex DotNetBasePathRegex = new Regex("Base Path:(.*)$", RegexOptions.Multiline);
         private static readonly Regex VersionRegex = new Regex(@"^(\d+)\.(\d+)\.(\d+)", RegexOptions.Multiline);
-        private static readonly Regex SdkRegex = new Regex(@"(\S+) \[(.*)]$", RegexOptions.Multiline);
+        private static readonly Regex SdkRegex = new Regex(@"(\S+) \[(.*?)]$", RegexOptions.Multiline);
 
         public static VisualStudioInstance GetInstance(string dotNetSdkPath)
         {            
--- a/src/MSBuildLocator/DotNetSdkLocationHelper.cs
+++ b/src/MSBuildLocator/DotNetSdkLocationHelper.cs
@@ -15,7 +15,7 @@ namespace Microsoft.Build.Locator
     {
         private static readonly Regex DotNetBasePathRegex = new Regex("Base Path:(.*)$", RegexOptions.Multiline);
         private static readonly Regex VersionRegex = new Regex(@"^(\d+)\.(\d+)\.(\d+)", RegexOptions.Multiline);
-        private static readonly Regex SdkRegex = new Regex(@"(\d+\.\d+\.\d+) \[(.*)]$", RegexOptions.Multiline);
+        private static readonly Regex SdkRegex = new Regex(@"(\S+) \[(.*)]$", RegexOptions.Multiline);
 
         public static VisualStudioInstance GetInstance(string dotNetSdkPath)
         {            
--- a/src/MSBuildLocator/DotNetSdkLocationHelper.cs
+++ b/src/MSBuildLocator/DotNetSdkLocationHelper.cs
@@ -96,7 +96,7 @@ namespace Microsoft.Build.Locator
             }
 
             if (process.HasExited)
-            {               
+            {
                 yield break;
             }
 
@@ -117,7 +117,7 @@ namespace Microsoft.Build.Locator
 
             var matched = DotNetBasePathRegex.Match(outputString);
             if (!matched.Success)
-            {                
+            {
                 yield break; 
             }
             
--- a/src/MSBuildLocator/DotNetSdkLocationHelper.cs
+++ b/src/MSBuildLocator/DotNetSdkLocationHelper.cs
@@ -12,12 +12,11 @@ using System.Text.RegularExpressions;
 namespace Microsoft.Build.Locator
 {
     internal static class DotNetSdkLocationHelper
-    {
-        private static readonly Regex DotNetBasePathRegex = new Regex("Base Path:(.*)$", RegexOptions.Multiline);
+    {        
         private static readonly Regex VersionRegex = new Regex(@"^(\d+)\.(\d+)\.(\d+)", RegexOptions.Multiline);
         private static readonly Regex SdkRegex = new Regex(@"(\d+\.\d+\.\d+) \[(.*)]$", RegexOptions.Multiline);
 
-        public static VisualStudioInstance GetInstance(string workingDirectory, string dotNetSdkPath)
+        public static VisualStudioInstance GetInstance(string dotNetSdkPath)
         {            
             if (string.IsNullOrWhiteSpace(dotNetSdkPath))
             {
@@ -63,7 +62,9 @@ namespace Microsoft.Build.Locator
 
             foreach (var basePath in basePaths)
             {
-                yield return GetInstance(workingDirectory, basePath);
+                var dotnetSdk = GetInstance(basePath);
+                if (dotnetSdk != null)
+                    yield return dotnetSdk;
             }
         }
 
@@ -76,7 +77,7 @@ namespace Microsoft.Build.Locator
             Process process;
             try
             {
-                var startInfo = new ProcessStartInfo("dotnet", "--info")
+                var startInfo = new ProcessStartInfo("dotnet", "--list-sdks")
                 {
                     WorkingDirectory = workingDirectory,
                     CreateNoWindow = true,
@@ -117,37 +118,22 @@ namespace Microsoft.Build.Locator
 
             var outputString = string.Join(Environment.NewLine, lines);
 
-            var matched = DotNetBasePathRegex.Match(outputString);
-            if (!matched.Success)
+            foreach (var line in lines)
             {
-                return null;
+                var sdkMatch = SdkRegex.Match(line);
+
+                if (!sdkMatch.Success)
+                    break;
+
+                var version = sdkMatch.Groups[1].Value.Trim();
+                var path = sdkMatch.Groups[2].Value.Trim();
+
+                path = Path.Combine(path, version) + Path.DirectorySeparatorChar;
+
+                basePaths.Add(path);
             }
 
-            var lineSdkIndex = lines.FindIndex(line => line.Contains(".NET Core SDKs installed") || line.Contains(".NET SDKs installed"));
-
-            if (lineSdkIndex != -1)
-            {
-                lineSdkIndex++; 
-
-                while (lineSdkIndex < lines.Count && (!lines[lineSdkIndex].Contains(".NET Core runtimes installed") && !lines[lineSdkIndex].Contains(".NET runtimes installed")))
-                {
-                    var sdkMatch = SdkRegex.Match(lines[lineSdkIndex]);
-
-                    if (!sdkMatch.Success)
-                        break;
-
-                    var version = sdkMatch.Groups[1].Value.Trim();                     
-                    var path = sdkMatch.Groups[2].Value.Trim();
-
-                    path = Path.Combine(path, version) + "\\";                    
-
-                    // We insert at index 0 so that instance list will be sorted descending so that instances.FirstOrDefault() 
-                    // will always return the latest installed version of dotnet SDK 
-                    basePaths.Insert(0, path);      
-
-                    lineSdkIndex++;
-                }
-            }            
+            basePaths.Reverse(); // Reverse the list in order to preserve FirstOrDefault to always return the latest installed SDK for the time being
             
             return basePaths;
         }
--- a/src/MSBuildLocator/DotNetSdkLocationHelper.cs
+++ b/src/MSBuildLocator/DotNetSdkLocationHelper.cs
@@ -142,10 +142,9 @@ namespace Microsoft.Build.Locator
 
                     var version = sdkMatch.Groups[1].Value.Trim();
                     var path = sdkMatch.Groups[2].Value.Trim();
-                    var test = Path.Combine(path, version); 
-                    //path = Path.Combine(path, version) + Path.DirectorySeparatorChar;
-                    path = Path.Combine(path, version);
-
+                    
+                    path = Path.Combine(path, version) + Path.DirectorySeparatorChar;
+                    
                     if (path.Equals(basePath))
                     {
                         // We insert the version in use at the front of the list in order to ensure FirstOrDefault always returns the version in use.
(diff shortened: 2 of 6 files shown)


samples/BuilderApp/BuilderApp.csproj
```diff
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFrameworks>net471;netcoreapp2.1</TargetFrameworks>
+    <TargetFrameworks>net472;netcoreapp2.1</TargetFrameworks>
```
JoeRobich: Is this a necessary change?

src/MSBuildLocator/DotNetSdkLocationHelper.cs
```diff
@@ -106,13 +117,44 @@ private static string GetDotNetBasePath(string workingDirectory)
...
+                    // We insert at index 0 so that instance list will be sorted descending so that instances.FirstOrDefault() 
+                    // will always return the latest installed version of dotnet SDK 
+                    basePaths.Insert(0, path);      
+
+                    lineSdkIndex++;
+                }
+            }
+
             var matched = DotNetBasePathRegex.Match(outputString);
             if (!matched.Success)
```
JoeRobich: Move this check before parsing sdks, since it would keep us from doing extra work.

src/MSBuildLocator/DotNetSdkLocationHelper.cs
```diff
@@ -106,13 +117,44 @@ private static string GetDotNetBasePath(string workingDirectory)
...
+
+                    if (!ma.Success)
+                        break;
+
+                    var version = ma.Groups[1].Value.Trim();                     
+                    var path = ma.Groups[2].Value.Trim();
+
+                    path = Path.Combine(path, version) + "\\";
+
+                    //basePaths.Add(path);
```
JoeRobich: Remove commented out code.

src/MSBuildLocator/DotNetSdkLocationHelper.cs
```diff
@@ -106,13 +117,44 @@ private static string GetDotNetBasePath(string workingDirectory)
...
 
+            var lineSdkIndex = lines.FindIndex(line => line.Contains(".NET Core SDKs installed"));
+
+            if (lineSdkIndex != -1)
+            {
+                lineSdkIndex++; 
+
+                while (lineSdkIndex < lines.Count && !lines[lineSdkIndex].Contains(".NET Core runtimes installed"))
+                {
+                    var ma = SdkRegex.Match(lines[lineSdkIndex]);
```
JoeRobich: Use a variable name like `sdkMatch` to provide more meaning.

src/MSBuildLocator/MSBuildLocator.cs
```diff
@@ -310,15 +310,14 @@ private static IEnumerable<VisualStudioInstance> GetInstances(VisualStudioInstan
             if (devConsole != null)
                 yield return devConsole;
 
-    #if FEATURE_VISUALSTUDIOSETUP
+#if FEATURE_VISUALSTUDIOSETUP
```
JoeRobich: Revert these indentation changes.

samples/BuilderApp/BuilderApp.csproj
kurtcodemander: No, fixed.

src/MSBuildLocator/DotNetSdkLocationHelper.cs
kurtcodemander: Fixed.

src/MSBuildLocator/DotNetSdkLocationHelper.cs
kurtcodemander: Fixed.

src/MSBuildLocator/DotNetSdkLocationHelper.cs
kurtcodemander: Fixed.

src/MSBuildLocator/MSBuildLocator.cs
kurtcodemander: Fixed.

kurtcodemander:
I have finally added another commit to this PR. I also added support for returning .NET 5.0 SDKs, which this PR did not previously have support for.

src/MSBuildLocator/DotNetSdkLocationHelper.cs
```diff
@@ -58,10 +57,22 @@ public static VisualStudioInstance GetInstance(string workingDirectory)
...
         }
 
-        private static string GetDotNetBasePath(string workingDirectory)
+        public static IEnumerable<VisualStudioInstance> GetInstances(string workingDirectory)
+        {
+            var basePaths = GetDotNetBasePaths(workingDirectory);
+
+            foreach (var basePath in basePaths)
+            {
+                yield return GetInstance(workingDirectory, basePath);
```
ladipro, line 66: nit: It would be cleaner to filter out `null` instances here and revert the extra check in `QueryVisualStudioInstances`.

src/MSBuildLocator/MSBuildLocator.cs
```diff
@@ -22,9 +22,12 @@ public static class MSBuildLocator
         private static readonly string[] s_msBuildAssemblies =
         {
             "Microsoft.Build",
+            "Microsoft.Build.Engine",
             "Microsoft.Build.Framework",
             "Microsoft.Build.Tasks.Core",
-            "Microsoft.Build.Utilities.Core"
+            "Microsoft.Build.Utilities.Core",
+            "System.Runtime.CompilerServices.Unsafe",
+            "System.Numerics.Vectors"
```
ladipro, line 30: These changes are already in the master branch, please rebase.

src/MSBuildLocator/DotNetSdkLocationHelper.cs
```diff
@@ -15,11 +15,10 @@ internal static class DotNetSdkLocationHelper
     {
         private static readonly Regex DotNetBasePathRegex = new Regex("Base Path:(.*)$", RegexOptions.Multiline);
         private static readonly Regex VersionRegex = new Regex(@"^(\d+)\.(\d+)\.(\d+)", RegexOptions.Multiline);
+        private static readonly Regex SdkRegex = new Regex(@"(\d+\.\d+\.\d+) \[(.*)]$", RegexOptions.Multiline);
 
-        public static VisualStudioInstance GetInstance(string workingDirectory)
-        {
-            string dotNetSdkPath = GetDotNetBasePath(workingDirectory);
-
+        public static VisualStudioInstance GetInstance(string workingDirectory, string dotNetSdkPath)
```
ladipro, line 20: `workingDirectory` is now unused and can be removed.

src/MSBuildLocator/DotNetSdkLocationHelper.cs
```diff
@@ -112,7 +123,33 @@ private static string GetDotNetBasePath(string workingDirectory)
                 return null;
             }
 
-            return matched.Groups[1].Value.Trim();
+            var lineSdkIndex = lines.FindIndex(line => line.Contains(".NET Core SDKs installed") || line.Contains(".NET SDKs installed"));
```
ladipro, line 126: Suggestion: Run `dotnet --list-sdks` instead of `dotnet --info` and you wouldn't have to match on complex human readable strings.

src/MSBuildLocator/DotNetSdkLocationHelper.cs
```diff
@@ -112,7 +123,33 @@ private static string GetDotNetBasePath(string workingDirectory)
...
+                {
+                    var sdkMatch = SdkRegex.Match(lines[lineSdkIndex]);
+
+                    if (!sdkMatch.Success)
+                        break;
+
+                    var version = sdkMatch.Groups[1].Value.Trim();                     
+                    var path = sdkMatch.Groups[2].Value.Trim();
+
+                    path = Path.Combine(path, version) + "\\";
```
ladipro, line 142: `Path.PathSeparator` instead of `\` for portability?

src/MSBuildLocator/DotNetSdkLocationHelper.cs
```diff
@@ -112,7 +123,33 @@ private static string GetDotNetBasePath(string workingDirectory)
...
+                        break;
+
+                    var version = sdkMatch.Groups[1].Value.Trim();                     
+                    var path = sdkMatch.Groups[2].Value.Trim();
+
+                    path = Path.Combine(path, version) + "\\";                    
+
+                    // We insert at index 0 so that instance list will be sorted descending so that instances.FirstOrDefault() 
+                    // will always return the latest installed version of dotnet SDK 
+                    basePaths.Insert(0, path);
```
ladipro, line 146: I don't see `dotnet` returning SDKs in a particular order documented anywhere. Also, this is O(N^2) and although not a big deal given the tiny number of items, it would be nicer to avoid it. My suggestion would be to return in the order as printed by `dotnet`, i.e. no sorting.

ladipro (changes requested):
Thank you @kurtcodemander! I left a few comments inline, mostly nits but please consider switching from `dotnet --info` to `dotnet --list-sdks`.

I believe that in addition to this change, we should also modify `RegisterDefaults()` to register the highest .NET SDK version **lower or equal** to the runtime of the executing process. This is to fully address https://github.com/microsoft/MSBuildLocator/issues/96. It will probably be a separate PR building on top of your work but I wanted leave a note since #96 seems to have broken many developers and should be treated as high-pri.

src/MSBuildLocator/DotNetSdkLocationHelper.cs
ladipro: You're correct though that if we don't add any smarts to this PR to pick the "right" SDK, we should keep the current behavior of returning the newest first. Maybe use `List<T>.Reverse()` then?

Forgind: For what it's worth, I don't think it's guaranteed that a foreach loop goes in order either. I don't think they'll ever change that unless they have aggressive parallelism, but it could be an argument for either explicitly sorting them or iterating (with a normal for loop) from top to bottom.

src/MSBuildLocator/MSBuildLocator.cs
```diff
@@ -81,7 +84,7 @@ public static IEnumerable<VisualStudioInstance> QueryVisualStudioInstances()
             IEnumerable<VisualStudioInstance> instances,
             VisualStudioInstanceQueryOptions options)
         {
-            return instances.Where(i => options.DiscoveryTypes.HasFlag(i.DiscoveryType));
+            return instances.Where(i => i != null && options.DiscoveryTypes.HasFlag(i.DiscoveryType));
```
Forgind, line 87: When is i null?

JoeRobich:
So the current behavior is that for the current working directory you will get back the newest SDK or the one pinned in the global.json, I would prefer it if this behavior is retained when invoking `MSBuildLocator.RegisterDefaults()`.

src/MSBuildLocator/MSBuildLocator.cs
ladipro: `GetInstance` can return `null` and consequently `GetInstances` can enumerate `null`. I, too, find it confusing. It would make more sense to not enumerate missing/broken instances in the first place.

src/MSBuildLocator/DotNetSdkLocationHelper.cs
kurtcodemander: Fixed

src/MSBuildLocator/MSBuildLocator.cs
kurtcodemander: Forgind merged master into this branch yesterday and I pulled those changes down today before committig the latest changes.

src/MSBuildLocator/DotNetSdkLocationHelper.cs
kurtcodemander: It has been removed.

src/MSBuildLocator/DotNetSdkLocationHelper.cs
kurtcodemander: Fixed. It now uses --list-sdks.

src/MSBuildLocator/DotNetSdkLocationHelper.cs
kurtcodemander: Fixed

src/MSBuildLocator/DotNetSdkLocationHelper.cs
kurtcodemander: The code now uses --list-sdks with a foreach loop and .Reverse. Need for any change to that?

src/MSBuildLocator/MSBuildLocator.cs
kurtcodemander: null check has been moved to GetInstances

kurtcodemander:
> So the current behavior is that for the current working directory you will get back the newest SDK or the one pinned in the global.json, I would prefer it if this behavior is retained when invoking `MSBuildLocator.RegisterDefaults()`.

It was suggested to use --list-sdks to populate the list of instances, which made sense to me, so that's what the latest commit contains, but it dawns on me that if there's one pinned in global.json then we'll probably have to use the "Base Path" from "dotnet --info" like before in order for the current behavior to be retained when invoking `MSBuildLocator.RegisterDefaults()`.

With the behavior from the latest commit https://github.com/microsoft/MSBuildLocator/pull/79/commits/b6353cb6a96dff279c2a72f4ab0a37f829f7a731 `MSBuildLocator.RegisterDefaults()` will always register the latest version of the SDK regardless of what's pinned in global.json. 

So, do we revert back to using dotnet --info again and checking the Base Path in addition to parsing SDKs from dotnet --info? Is there an argument for invoking dotnet.exe twice in order to parse Base Path from dotnet --info and then the list of SDKs from dotnet --list-sdks? The latter doesn't feel quite right.

Forgind (commented):
dotnet --version should be a more efficient/simpler way to get the pinned version. I tried putting `--version & dotnet --list-sdks` into the arguments, but that didn't work. I like the simplicity of calling dotnet --list-sdks for all the sdks and dotnet --version for the pinned one if relevant, but I'd prefer not to start two processes if it's avoidable. Not sure what's best.

src/MSBuildLocator/DotNetSdkLocationHelper.cs
```diff
@@ -106,13 +118,24 @@ private static string GetDotNetBasePath(string workingDirectory)
...
-                return null;
+                var sdkMatch = SdkRegex.Match(line);
+
+                if (!sdkMatch.Success)
+                    break;
+
+                var version = sdkMatch.Groups[1].Value.Trim();
+                var path = sdkMatch.Groups[2].Value.Trim();
+
+                path = Path.Combine(path, version) + Path.DirectorySeparatorChar;
```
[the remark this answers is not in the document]
kurtcodemander: Path.Combine doesn't add the directory separator to the end so in order to preserve VisualStudioInstance.VisualStudioRootPath and VisualStudioRootPath.MSBuildPath having a trailing separator then we need it.

kurtcodemander:
Since it doesn't seem to be possible to get both the pinned version in use and the list of SDKs starting only one process without using dotnet --info I have made another commit going back to dotnet --info, but also making some improvements to the older version which used --info. 

Now comparing each SDK path to the base path of the version in use, inserting the version in use at the top of the list whilst simply adding the rest to the list. The `Reverse()` have been removed since it's no longer needed. 

Now, we preserve the functionality of FirstOrDefault() always returning the pinned version in use as reflected by global.json. 

A separate PR should probably address the [comment ](https://github.com/microsoft/MSBuildLocator/pull/79#pullrequestreview-539945239)from @ladipro and #96 

Comments on the latest commit and this way of doing it?

Forgind (approved):
Looks good to me, thanks!

src/MSBuildLocator/DotNetSdkLocationHelper.cs
```diff
@@ -111,8 +124,42 @@ private static string GetDotNetBasePath(string workingDirectory)
...
+                {
+                    var sdkMatch = SdkRegex.Match(lines[lineSdkIndex]);
+
+                    if (!sdkMatch.Success)
+                        break;
+
+                    var version = sdkMatch.Groups[1].Value.Trim();
+                    var path = sdkMatch.Groups[2].Value.Trim();
+
+                    path = Path.Combine(path, version) + Path.DirectorySeparatorChar;
```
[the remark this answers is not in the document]
kurtcodemander: I see, but at the bottom in GetInstance we set dotNetSdkPath directly into new VisualStudioInstance(..) and the VisualStudioInstance constructor saves it directly to VisualStudioRootPath and MSBuildPath so that's where it would break the existing functionality having a trailing separator.

src/MSBuildLocator/DotNetSdkLocationHelper.cs
```diff
@@ -111,8 +124,42 @@ private static string GetDotNetBasePath(string workingDirectory)
...
+
+                    if (path.Equals(basePath))
+					{
+                        // We insert the version in use at the top of the list in order to preserve FirstOrDefault to always return the version in use
```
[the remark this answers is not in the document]
kurtcodemander: Looks good, fixed

src/MSBuildLocator/DotNetSdkLocationHelper.cs
```diff
@@ -111,8 +124,42 @@ private static string GetDotNetBasePath(string workingDirectory)
...
+                    if (!sdkMatch.Success)
+                        break;
+
+                    var version = sdkMatch.Groups[1].Value.Trim();
+                    var path = sdkMatch.Groups[2].Value.Trim();
+
+                    path = Path.Combine(path, version) + Path.DirectorySeparatorChar;
+
+                    if (path.Equals(basePath))
+					{
```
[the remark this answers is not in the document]
kurtcodemander: Fixed

src/MSBuildLocator/DotNetSdkLocationHelper.cs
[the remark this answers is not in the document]
Forgind: Ah, missed that. You're right.

src/MSBuildLocator/DotNetSdkLocationHelper.cs
```diff
@@ -15,11 +15,10 @@ internal static class DotNetSdkLocationHelper
     {
         private static readonly Regex DotNetBasePathRegex = new Regex("Base Path:(.*)$", RegexOptions.Multiline);
         private static readonly Regex VersionRegex = new Regex(@"^(\d+)\.(\d+)\.(\d+)", RegexOptions.Multiline);
+        private static readonly Regex SdkRegex = new Regex(@"(\d+\.\d+\.\d+) \[(.*)]$", RegexOptions.Multiline);
```
ladipro, line 18: This won't work for rc's, previews, and other releases with non-numeric version strings. For example:

```
.NET SDKs installed:
5.0.100-preview.8.20373.19 [C:\Program Files\dotnet\sdk]
```

The regex should be made more permissive.

ladipro (changes requested):
@kurtcodemander thank you for following up and apologies for having randomized you with `--list-sdks`. I am requesting changes to fix an issue with the SDK version regex, please see inline. Other than that it looks good to me.

src/MSBuildLocator/DotNetSdkLocationHelper.cs
kurtcodemander: Any suggestion for a change to the regex to make it more permissive? RegEx is one of my achilles' heels :-p

ladipro: I'd try ```(\S+) \[(.*)]$```, i.e. expecting the version to be a non-empty string of non-whitespace characters.

Forgind: I tried that with "14.3.155-preview-02334alpha", and it didn't seem to work. I think \S is whitespace rather than non-whitespace characters (fixable with `^`?) and it also makes it harder to parse out groups.

I used @"(\d+)\.(\d+)\.(\d+)", and that seemed to work pretty well.

ladipro: Maybe it appears to work because it's not matching the entire version. Based on this fiddle `(\d+)\.(\d+)\.(\d+)` is not working and `\S+' looks OK: https://dotnetfiddle.net/QS5kui

Forgind: Interesting...I was testing with a local ConsoleApp that didn't agree. Did I do something wrong?
image

I tweaked my version to make it work: https://dotnetfiddle.net/0IUKbB

ladipro: In your screenshot `toMatch` is just a version, it does not have the space and square brackets as required by the regex.

Forgind: Ah, gotcha. Works now. 👍

kurtcodemander: I have commited this change now.

src/MSBuildLocator/DotNetSdkLocationHelper.cs
```diff
@@ -111,8 +124,42 @@ private static string GetDotNetBasePath(string workingDirectory)
             {
                 return null;
             }
+            
+            var basePath = matched.Groups[1].Value.Trim();
```
ladipro, line 128: nit: `GetDotNetBasePaths` could be an iterator method and simply `yield return basePath` here and then `yield return path` in the loop below. Just a thought, looks good either way.

ladipro (approved):
Thank you!

src/MSBuildLocator/DotNetSdkLocationHelper.cs
kurtcodemander: I like this idea and have submitted a new commit with this change. Looks quite nice and clean.

src/MSBuildLocator/DotNetSdkLocationHelper.cs
```diff
@@ -109,10 +118,37 @@ private static string GetDotNetBasePath(string workingDirectory)
...
+                    if (!sdkMatch.Success)
+                        break;
+
+                    var version = sdkMatch.Groups[1].Value.Trim();
+                    var path = sdkMatch.Groups[2].Value.Trim();
+                    
+                    path = Path.Combine(path, version) + Path.DirectorySeparatorChar;
+
+                    if (!path.Equals(basePath))
+                        yield return path;
```
rainersigwald, line 147: Part of me wants to order this list by version descending (`dotnet --info` returns results in ascending order). But we already extract the `global.json`-or-highest version as `basePath` first, and that ordering is most critical. And we don't order VS instances. So I think the current approach is good.

rainersigwald (commented):
Thanks! Note that we have a bunch of PRs to land and this adds a new feature so it might need to sit a bit while we work through some bugfixes.

rainersigwald:
Also, related SDK feature request: dotnet/sdk#9865. Does no good yet, of course.

src/MSBuildLocator/DotNetSdkLocationHelper.cs
Forgind: I think it's still useful to sort the results even with the basePath first because of https://github.com/microsoft/MSBuildLocator/pull/106. If the default SDK is from after the runtime in use, it'll be disqualified, and we'd go with the second in the list. If you don't clean your machine regularly and still have a 1.0 SDK lying around, that's what we'd find and register with RegisterDefaults.

rainersigwald: I'm convinced.

kurtcodemander: So should I remove the new iterator/yield code and go back to the code using List<string> with Reverse(), or do we have a better solution based on the current code?

Forgind: I think that's the best solution. I already wrote up the change in #106, so I wouldn't worry about remaking that change here.

rainersigwald (approved):
Thanks @kurtcodemander!

PR #82, opened 2020-01-17
Add support for Mono

radical:
- on OSX, we get the version from the path
	- we return the version of currently running mono, as the first
	result
         - followed by mono with versions from `mono --version=number`
- on linux/windows, we try to get the version from `mono --version=number`
	- if that fails, then we return the instance with `0.0.0.0`


src/MSBuildLocator/MSBuildLocator.cs
```diff
@@ -323,6 +339,87 @@ private static IEnumerable<VisualStudioInstance> GetInstances(VisualStudioInstan
...
+                yield return new VisualStudioInstance("Mono", runningMonoPath, version, DiscoveryType.Mono);
+            }
+
+            if (!IsOSX)
+            {
+                // Returning just one instance on !osx
+                yield break;
+            }
+
+            foreach(var dirPath in Directory.EnumerateDirectories(s_monoOSXBasePath))
```
akoeplinger: do we really needs this? `mono --version=number` should work fine on OSX.

src/MSBuildLocator/MSBuildLocator.cs
```diff
@@ -354,5 +451,53 @@ private static VisualStudioInstance GetDevConsoleInstance()
...
+        private static bool? _isRunningOnMono;
+
+        /// <summary>
+        /// Gets a flag indicating if we are running under MONO
+        /// </summary>
+        internal static bool IsRunningOnMono
+        {
+            get
+            {
+                if (_isRunningOnMono != null) return _isRunningOnMono.Value;
```
akoeplinger: nit:
```suggestion
                if (_isRunningOnMono.HasValue) return _isRunningOnMono.Value;
```

src/MSBuildLocator/MSBuildLocator.cs
```diff
@@ -354,5 +451,53 @@ private static VisualStudioInstance GetDevConsoleInstance()
...
+
+                lock (IsRunningOnMonoLock)
+                {
+                    if (_isRunningOnMono == null)
+                    {
+                        // There could be potentially expensive TypeResolve events, so cache IsMono.
+                        // Also, VS does not host Mono runtimes, so turn IsMono off when msbuild is running under VS
```
akoeplinger: I assume you're just talking about VS on Windows right?

radical: yep

src/MSBuildLocator/MSBuildLocator.cs
radical: this and the `IsRunningOnMono` are as-is from msbuild. I'll address the feedback though.

src/MSBuildLocator/MSBuildLocator.cs
radical: Switched to using `mono --version=number`. On OSX, doing this for all the versions found.

radical:
@rainersigwald ping

radical:
@rainersigwald while I have your attention, ping ;)

src/MSBuildLocator/DiscoveryType.cs
```diff
@@ -25,6 +25,8 @@ public enum DiscoveryType
         /// <summary>
         ///     Discovery via dotnet --info.
         /// </summary>
-        DotNetSdk = 4
+        DotNetSdk = 4,
+
+        Mono = 8
```
JoeRobich, line 30: doc comment
```suggestion
        /// <summary>
        ///     Discovery via Mono path testing.
        /// </summary>
        Mono = 8
```

src/MSBuildLocator/MSBuildLocator.cs
```diff
@@ -323,6 +338,85 @@ private static IEnumerable<VisualStudioInstance> GetInstances(VisualStudioInstan
...
+                    return Version.TryParse(stdout_str, out ver);
+                } catch (Win32Exception) {
+                }
+
+                return false;
+            }
+        }
+
+
+
```
JoeRobich, line 419: nit: consecutive blank newlines

src/MSBuildLocator/MSBuildLocator.cs
```diff
@@ -354,5 +448,52 @@ private static VisualStudioInstance GetDevConsoleInstance()
...
+            {
+                if (!_isOSX.HasValue)
+                {
+                    _isOSX = File.Exists("/usr/lib/libc.dylib");
+                }
+
+                return _isOSX.Value;
+            }
+        }
+
```
JoeRobich, line 497: nit: blank newline before closing brace
```suggestion
```

src/MSBuildLocator/MSBuildLocator.cs
```diff
@@ -323,6 +338,83 @@ private static IEnumerable<VisualStudioInstance> GetInstances(VisualStudioInstan
...
+            if (!IsOSX)
+            {
+                // Returning just one instance on !osx
+                yield break;
+            }
+
+            foreach(var dirPath in Directory.EnumerateDirectories(s_monoOSXBasePath))
+            {
+                if (string.Compare(Path.GetFileName (dirPath), "Current") == 0 || // skip the `Current` symlink
+                    string.Compare(dirPath, runningMonoPath) == 0)               // and the running mono version
```
Forgind, line 359: nit:
```suggestion
                if (string.Equals(Path.GetFileName(dirPath), "Current") || // skip the `Current` symlink
                    string.Equals(dirPath, runningMonoPath))                   // and the running mono version
```

Also, this feels a little fragile because it would fail if someone appends a slash to dirPath. That may sound arbitrary, but other places we make a new VisualStudioInstance append a slash to the end in setting the path, and people can access that directly using QueryVisualStudioInstances, so not having that would be a confusing difference in how it works.

Forgind (changes requested):
Would you mind moving the changes to find mono into a separate class/file?

Forgind:
Also, per https://github.com/microsoft/MSBuildLocator/pull/115#discussion_r551476516, would you mind pulling in that PR and adding the check?

src/MSBuildLocator/MSBuildLocator.cs
radical: Append a `/` to the directory name itself? `dirPath` is from `Directory.EnumerateDirectories`

Forgind: Right. The two scenarios I'm talking about here are:
1) Someone expects dirPath to end in a slash because it does for other runtimes. They have code that uses that --> error.
2) Someone notes that dirPath ends in a slash for other runtimes and decides to centralize it with some method that's called before this. Slash is added to dirPath --> Path.GetFileName returns an empty string --> this part fails.

The first is a current problem with this code, and the second is a hypothetical I'd like to avoid. Is that clearer?

radical: Ok. Changed it to `string.Equals(Path.GetFullPath(dirPath), runningMonoFullPath)`

src/MSBuildLocator/MSBuildLocator.Mono.cs
```diff
@@ -0,0 +1,144 @@
+using System;
```
Forgind, line 1: Copyright header

src/MSBuildLocator/MSBuildLocator.Mono.cs
```diff
@@ -0,0 +1,144 @@
...
+                if (TryGetValidMonoVersion(dirPath, out version))
+                {
+                    yield return new VisualStudioInstance("Mono", dirPath, version, DiscoveryType.Mono);
+                }
+            }
+
+            bool TryGetValidMonoVersion(string path, out Version ver)
+            {
+                ver = null;
+                if (!File.Exists(Path.Combine(path, s_monoMSBuildDll_Current_RelativePath)) &&
```
Forgind, line 47: tiny nit: indent less to line them up

src/MSBuildLocator/MSBuildLocator.Mono.cs
```diff
@@ -0,0 +1,144 @@
...
+            {
+                if (string.Equals(Path.GetFileName(dirPath), "Current") || // skip the `Current` symlink
+                    string.Equals(Path.GetFullPath(dirPath), runningMonoFullPath)) // and the running mono version
+                {
+                    continue;
+                }
+
+                if (TryGetValidMonoVersion(dirPath, out version))
+                {
+                    yield return new VisualStudioInstance("Mono", dirPath, version, DiscoveryType.Mono);
```
Forgind, line 40:
```suggestion
                    yield return new VisualStudioInstance("Mono", dirPath + '/', version, DiscoveryType.Mono);
```

src/MSBuildLocator/MSBuildLocator.Mono.cs
```diff
@@ -0,0 +1,144 @@
...
+                }
+
+                if (TryGetMonoVersionFromMonoBinary(path, out ver) || Version.TryParse(Path.GetFileName(path), out ver))
+                {
+                    return true;
+                }
+
+                // The path has a valid mono, but we can't find the version
+                // so, let's return the instance at least but with version 0.0.0
+                ver = new Version(0, 0, 0);
```
Forgind, line 60: Not sure whether this is better or it's better to not return it. Users could find it separately and register it anyway. @rainersigwald, thoughts?

Forgind (approved):
Also remember https://github.com/microsoft/MSBuildLocator/pull/82#issuecomment-754129837, but I think this is good!

Forgind (approved):

src/MSBuildLocator/MSBuildLocator.Mono.cs
radical: Ok to merge? Or are we waiting for this?

Forgind: I'd prefer to wait for @rainersigwald's signoff. Also, would you mind adding the if-not-mono check discussed [here](https://github.com/microsoft/MSBuildLocator/pull/82#issuecomment-754129837)?

radical: Added, and fixed a bug introduced in the nuget PR.

src/MSBuildLocator/MSBuildLocator.Mono.cs
```diff
@@ -0,0 +1,147 @@
...
+        /// <summary>
+        /// Gets a flag indicating if we are running under Mac OSX
+        /// </summary>
+        internal static bool IsOSX
+        {
+            get
+            {
+                if (!_isOSX.HasValue)
+                {
+                    _isOSX = File.Exists("/usr/lib/libc.dylib");
```
rainersigwald, line 140: Doesn't this break on Big Sur? Can you call `RuntimeInformation.IsOSPlatform` instead? And in that case it might be worth just inlining the call instead of caching it here.

src/MSBuildLocator/MSBuildLocator.Mono.cs
rainersigwald: This is probably ok, let's just listen for feedback.

rainersigwald (commented):
Overall LGTM. Thanks for your patience on this!

rainersigwald (approved):

radical:
Does `<PackageReference Include="System.Runtime.InteropServices.RuntimeInformation" Version="4.3.0" />` look correct?

src/MSBuildLocator/Microsoft.Build.Locator.csproj
```diff
@@ -26,6 +26,7 @@
   </ItemGroup>
 
   <ItemGroup>
+    <PackageReference Include="System.Runtime.InteropServices.RuntimeInformation" Version="4.3.0" />
```
rainersigwald, line 29: Ah, this is unfortunate. I told you to do this! But . . . it's really nice that the Locator package has no dependencies so you can use it basically anywhere. Unfortunately, this type wasn't added to .NET Framework until 4.7.1, so this reference is required to call it in a straightforward way, which makes the package slightly less easy to use.

I think we can call it via reflection instead to avoid that hard requirement--if the type/method is not available, it's not macOS!

rainersigwald (changes requested):

PR #84, opened 2020-02-12
Add support for finding CoreXT instances

jeffkl:
CoreXT is a Microsoft internal-only build environment.  We need tools to work in all environments so adding support for CoreXT makes MSBuildLocator more useful.

### DIFF
--- a/src/MSBuildLocator.Tests/TemporaryEnvironmentVariables.cs
+++ b/src/MSBuildLocator.Tests/TemporaryEnvironmentVariables.cs
@@ -1,0 +1,43 @@
+﻿using System;
+using System.Collections.Generic;
+
+namespace Microsoft.Build.Locator.Tests
+{
+    /// <summary>
+    /// Represents a class for temporarily setting environment variables.
+    /// </summary>
+    public class TemporaryEnvironmentVariables : IDisposable
+    {
+        private readonly Dictionary<string, string> _backup = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="TemporaryEnvironmentVariables" /> class.
+        /// </summary>
+        /// <param name="environmentVariables">A <see cref="IDictionary{String,String}" /> containing the temporary environment variables to set.</param>
+        public TemporaryEnvironmentVariables(IDictionary<string, string> environmentVariables)
+        {
+            if (environmentVariables == null)
+            {
+                throw new ArgumentNullException(nameof(environmentVariables));
+            }
+
+            foreach (var environmentVariable in environmentVariables)
+            {
+                // Back up existing value
+                _backup[environmentVariable.Key] = Environment.GetEnvironmentVariable(environmentVariable.Key);
+
+                // Set new value
+                Environment.SetEnvironmentVariable(environmentVariable.Key, environmentVariable.Value);
+            }
+        }
+
+        public void Dispose()
+        {
+            foreach (var environmentVariable in _backup)
+            {
+                // Restore backed up value
+                Environment.SetEnvironmentVariable(environmentVariable.Key, environmentVariable.Value);
+            }
+        }
+    }
+}
--- a/src/MSBuildLocator/DiscoveryType.cs
+++ b/src/MSBuildLocator/DiscoveryType.cs
@@ -25,6 +25,11 @@ namespace Microsoft.Build.Locator
         /// <summary>
         ///     Discovery via dotnet --info.
         /// </summary>
-        DotNetSdk = 4
+        DotNetSdk = 4,
+
+        /// <summary>
+        /// Discovery via the Microsoft internal-only build environment CoreXT.
+        /// </summary>
+        CoreXT = 8
     }
 }
--- a/src/MSBuildLocator/DiscoveryType.cs
+++ b/src/MSBuildLocator/DiscoveryType.cs
@@ -30,6 +30,6 @@ namespace Microsoft.Build.Locator
         /// <summary>
         /// Discovery via the Microsoft internal-only build environment CoreXT.
         /// </summary>
-        CoreXT = 8
+        CoreXT = 8,
     }
 }
--- a/src/MSBuildLocator/MSBuildLocator.cs
+++ b/src/MSBuildLocator/MSBuildLocator.cs
@@ -349,7 +349,6 @@ namespace Microsoft.Build.Locator
 
             var msbuildPath = Environment.GetEnvironmentVariable($"MSBuildToolsPath_{toolset}");
 
-
             if (string.IsNullOrWhiteSpace(msbuildPath) || !(directoryExists ?? Directory.Exists)(msbuildPath))
             {
                 // A corresponding MSBuildToolsPath_XXX env var must be set and it must exist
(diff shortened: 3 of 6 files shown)


src/MSBuildLocator/DiscoveryType.cs
```diff
@@ -25,6 +25,11 @@ public enum DiscoveryType
         /// <summary>
         ///     Discovery via dotnet --info.
         /// </summary>
-        DotNetSdk = 4
+        DotNetSdk = 4,
+
+        /// <summary>
+        /// Discovery via the Microsoft internal-only build environment CoreXT.
+        /// </summary>
+        CoreXT = 8
```
rainersigwald:
```suggestion
        CoreXT = 8,
```

(this nit minimizes the next diff)

src/MSBuildLocator/MSBuildLocator.cs
```diff
@@ -324,6 +330,44 @@ private static IEnumerable<VisualStudioInstance> GetInstances(VisualStudioInstan
...
+            }
+
+            if (!Version.TryParse(Environment.GetEnvironmentVariable("VisualStudioVersion"), out Version visualStudioVersion))
+            {
+                // VisualStudioVersion env var must be set to a valid version
+                return null;
+            }
+
+            // Find the newest installed Visual Studio instance that corresponds with the VisualStudioVersion set by CoreXT
+            VisualStudioInstance visualStudioInstance = (visualStudioInstances ?? VisualStudioLocationHelper.GetInstances()).Where(i => i.Version.Major == visualStudioVersion.Major).OrderByDescending(i => i.Version).FirstOrDefault();
```
rainersigwald: I'm confused by this. I would have expected that in a CoreXT environment we'd load only the MSBuild.CoreXT-package delivered MSBuild; this instead loads . . . a hybrid of the CoreXT targets and the installed-VS engine?

jeffkl: As far as I can tell, the only property used when registering an instance is the MSBuildPath.  So this logic essentially lets callers know the path to Visual Studio in case they want to use it but the MSBuild path is the CoreXT one and that is the path used for loading assemblies.  Technically there is no Visual Studio path for the CoreXT environment so I thought this was a cool feature to associate the closest installed VS for the MSBuild.Corext package in use.

rainersigwald: This doesn't seem that useful to me. Can we not, and add it in later if we see a concrete use case?

src/MSBuildLocator/MSBuildLocator.cs
```diff
@@ -324,6 +330,44 @@ private static IEnumerable<VisualStudioInstance> GetInstances(VisualStudioInstan
...
+        /// <summary>
+        /// Gets CoreXT instances.
+        /// </summary>
+        /// <param name="visualStudioInstances">Optional <see cref="IList{VisualStudioInstance} "/> containing instances.  If none are specified, they are discovered through the Visual Studio Setup API.
+        /// This is only here for unit tests</param>
+        /// <param name="directoryExists">An optional <see cref="Func{String,Boolean} "/> to use when checking directory existence.  Should only be set by unit tests.</param>
+        /// <returns>A <see cref="VisualStudioInstance" /> object representing the Visual Studio instance if one is found, otherwise null.</returns>
+        internal static VisualStudioInstance GetCoreXTInstance(IList<VisualStudioInstance> visualStudioInstances = null, Func<string, bool> directoryExists = null)
+        {
+            var toolset = Environment.GetEnvironmentVariable("MsBuildToolset")?.Trim();
```
rainersigwald: Should we also have a check for a CoreXT environment variable like maybe `CoreXtExeVersion` to limit false positives?

src/MSBuildLocator/MSBuildLocator.cs
jeffkl: Without this, my primary use case I need it for has to query the instances separately.  This is very useful for me to replace all of this code:

https://github.com/microsoft/slngen/blob/master/src/Microsoft.VisualStudio.SlnGen/Program.Configure.cs#L26-L44

Of course this would just be a default, users could still manually find VS instances by calling `QueryVisualStudioInstances`.  But I feel that most callers would just want to know the best VS installation for the specified `MSBuild.Corext`.

jeffkl: I'm going to stop using `Microsoft.Build.Locator` for my application since I need to locate MSBuild and Visual Studio.  I'll remove the code that tries to find a VS instance for CoreXT and just keep this PR about finding only MSBuild.

src/MSBuildLocator.Tests/TemporaryEnvironmentVariables.cs
```diff
@@ -0,0 +1,43 @@
+﻿using System;
```
Forgind, line 1: copyright header

src/MSBuildLocator.Tests/TemporaryEnvironmentVariables.cs
```diff
@@ -0,0 +1,43 @@
...
+using System.Collections.Generic;
+
+namespace Microsoft.Build.Locator.Tests
+{
+    /// <summary>
+    /// Represents a class for temporarily setting environment variables.
+    /// </summary>
+    public class TemporaryEnvironmentVariables : IDisposable
```
Forgind, line 9: TestEnvironment?

src/MSBuildLocator/MSBuildLocator.cs
```diff
@@ -324,6 +330,41 @@ private static IEnumerable<VisualStudioInstance> GetInstances(VisualStudioInstan
...
+
+            if (string.IsNullOrWhiteSpace(toolset) || !int.TryParse(toolset, out int toolsetVersion) || toolsetVersion < 150)
+            {
+                // MsBuildToolset is not set or is less than 150
+                return null;
+            }
+
+            var msbuildPath = Environment.GetEnvironmentVariable($"MSBuildToolsPath_{toolset}");
```
Forgind, line 350: Does this normally end in a slash? Do we use CoreXT on case-sensitive OSs? Does `Environment.GetEnvironmentVariable` care?

src/MSBuildLocator/MSBuildLocator.cs
```diff
@@ -324,6 +330,41 @@ private static IEnumerable<VisualStudioInstance> GetInstances(VisualStudioInstan
...
+
+            if (string.IsNullOrWhiteSpace(msbuildPath) || !(directoryExists ?? Directory.Exists)(msbuildPath))
+            {
+                // A corresponding MSBuildToolsPath_XXX env var must be set and it must exist
+                return null;
+            }
+
+            if (!Version.TryParse(Environment.GetEnvironmentVariable("VisualStudioVersion"), out Version visualStudioVersion))
```
Forgind, line 358: This doesn't look like it would handle 1.6.8.0-preview-gobbledygook

src/MSBuildLocator/MSBuildLocator.cs
```diff
@@ -324,6 +330,41 @@ private static IEnumerable<VisualStudioInstance> GetInstances(VisualStudioInstan
...
 
 #if NET46
+        /// <summary>
+        /// Gets CoreXT instances.
+        /// </summary>
+        /// <param name="visualStudioInstances">Optional <see cref="IList{VisualStudioInstance} "/> containing instances.  If none are specified, they are discovered through the Visual Studio Setup API.
+        /// This is only here for unit tests</param>
+        /// <param name="directoryExists">An optional <see cref="Func{String,Boolean} "/> to use when checking directory existence.  Should only be set by unit tests.</param>
+        /// <returns>A <see cref="VisualStudioInstance" /> object representing the Visual Studio instance if one is found, otherwise null.</returns>
+        internal static VisualStudioInstance GetCoreXTInstance(IList<VisualStudioInstance> visualStudioInstances = null, Func<string, bool> directoryExists = null)
```
Forgind, line 340: Move this to its own class?

Issue #68, opened 2019-06-04
Query options and sorting


clairernovotny:
As far as I've seen, there's no way to sort or filter on VS options.

The following would be useful:

- Specify 2019 or 2017
- Preview or Stable
- Sort (ascending/descending, can be combined with preview/stable)

rainersigwald:
The philosophy so far has been to push this onto the calling code, allowing it to get a list of options and filter/sort that in its own way.

https://github.com/microsoft/MSBuildLocator/blob/8b5c58f5842f58ebf860ece96ac0f6924df839b1/samples/BuilderApp/Program.cs#L26-L27

Do you think there's a better API that we could provide? What would that look like?

onovotny:
The problem with "Ask what to use" is that it may not be interactive. 

From an API perspective, I think filters could be added to the query option. It would make it declarative, "give me 2019 or return emtpy if not found"

onovotny:
Another idea, which could be combined with the query option, is to add additional properties to the instance objects, for the version number and channel name.

rainersigwald:
I'd be on board with adding fields to our object to make filtering easier. Unfortunately, the things you're interested in don't seem to be available in [`ISetupInstance2`](https://docs.microsoft.com/en-us/dotnet/api/microsoft.visualstudio.setup.configuration.isetupinstance2).

Adding filtering options to the query doesn't strike me as better than letting the caller do it, though.

onovotny:
Does one of those at least indicate the channel name & stable/pre-release-ness? That'd be a good starting point.

rainersigwald:
Not that I can see. The only "properties" I see are:

```
campaignId: 1060713956.1557239266
setupEngineFilePath: C:\Program Files (x86)\Microsoft Visual Studio\Installer\vs_installershell.exe
nickname:
channelManifestId: VisualStudio.16.Release/16.1.1+28922.388
```

onovotny:
How does `vswhere` do it then, there must be a way? It has a `-prerelease` flag (and skips those by default).

jnm2:
The ability to filter to require workloads is important, too. Finding the wrong VS Build Tools installation was the cause of the `The SDK 'Microsoft.NET.Sdk' specified could not be found` failure for us, and I see a couple issues in this repo with that same message.

On a build server with:

- VS Build Tools 2019 with no workloads and only certain components installed
- VS Build Tools 2022 with .NET web and desktop workloads

[...]

PR #87, opened 2020-03-25
Add ProductId for Visual Studio instances

jeffkl:
Partially addresses https://github.com/microsoft/MSBuildLocator/issues/68
Helps me unblock https://github.com/microsoft/slngen/issues/121


jeffkl:
Hmm, maybe I should just add a new class that wraps ISetupInstance and returns everything.  And then our `VisualStudioInstance` object would have a `VisualStudioInfo` property that had it.  Thoughts?

jeffkl:
I'm going to abandon this PR and send a different revision that just plumbs through the whole VS instance object.

PR #94, opened 2020-08-27
Included additional assemblies to load on demand

EWSoftware:
Added Microsoft.Build.Engine and two supporting assemblies to the set that can be loaded.  These needed to be included in order for my project to work properly with the build locator.

### DIFF
--- a/src/MSBuildLocator/MSBuildLocator.cs
+++ b/src/MSBuildLocator/MSBuildLocator.cs
@@ -22,9 +22,12 @@ namespace Microsoft.Build.Locator
         private static readonly string[] s_msBuildAssemblies =
         {
             "Microsoft.Build",
+            "Microsoft.Build.Engine",
             "Microsoft.Build.Framework",
             "Microsoft.Build.Tasks.Core",
-            "Microsoft.Build.Utilities.Core"
+            "Microsoft.Build.Utilities.Core",
+            "System.Runtime.CompilerServices.Unsafe",
+            "System.Numerics.Vectors"
         };
 
 #if NET46


mattmurp:
We have a similar problem where we need Microsoft.Build.Engine.dll to be included in the list.  For the other 2 assemblies, I'm assuming that's related to https://github.com/microsoft/MSBuildLocator/issues/95.  We were able to use binding redirects as a work around for now.

EWSoftware:
In my case the two System assemblies aren't part of my project so there's nothing to redirect to.  I didn't feel I should have to add them to my project as they are already part of the MSBuild file set and it can find them there so I added them to the build locator.

ladipro (commented):
This is indeed an issue with the way assembly lookup currently works in MSBuildLocator. Using `System.Runtime.CompilerServices.Unsafe` as an example, the version shipped with the latest MSBuild is 4.0.6.0 while `System.Memory` references version 4.0.4.1. This version mismatch makes the default assembly load fail and requires a more explicit handling -- like matching on name + token only, as implemented in this PR.

Regular MSBuild execution via MSBuild.exe is fine because MSBuild.exe.config contains appropriate binding redirects, e.g.:

```xml
    <assemblyBinding xmlns="urn:schemas-microsoft-com:asm.v1">
      <dependentAssembly>
        <assemblyIdentity name="System.Runtime.CompilerServices.Unsafe" publicKeyToken="b03f5f7f11d50a3a" culture="neutral" />
        <bindingRedirect oldVersion="0.0.0.0-4.0.6.0" newVersion="4.0.6.0" />
      </dependentAssembly>
    </assemblyBinding>
```

I am unsure if this PR is a good way to address this issue, though. It would be be best if the list of MSBuild assemblies really contained only assemblies with publicly exposed MSBuild functionality because the list of dependencies possibly requiring binding redirects is an implementation detail and might be different in each new version of MSBuild.

Two alternative approaches I can think of:

1) Modify the assembly load handler to figure out assembly references (of the stable set of "root" MSBuild assemblies) and build the list of dependencies dynamically. Either by using information passed to the handler by CLR (requesting assembly?) or by scanning them manually with a metadata reader.

2) Modify the assembly load handler to open MSBuild.exe.config from the MSBuild installation path and emulate the redirects accordingly.

If we are fine with no forward compatibility guarantees, I guess we could take this PR and then make sure that the list is updated every time MSBuild dependencies change. But it looks too fragile indeed.

Wondering what others think, thanks!

ladipro:
For completeness, the workaround is to copy the relevant `<dependentAssembly>` tags from MSBuild.exe.config to the config file of the application hosting MSBuild. It has the same downsides as this fix, i.e. it is not future-proof.

yaakov-h:
@ladipro Copying the binding redirects only work when the consuming application also ships those assemblies, and the correct versions of them.

If I have an application that consumes Microsoft.Build.* but does not include System.Runtime.CompilerServices.Unsafe or System.Numerics.Vectors, then adding a binding redirect achieves nothing.

This will become more of a problem as Microsoft.Build continues to expands its dependency tree, as you have identified.

Requiring manual management of these redirects or redistributing these dependencies breaks the core contracts of MSBuildLocator, which is that the consumer can simply reference Microsoft.Build without redistributing it.

This PR is a very good interim fix, and though it doesn't fix the ever-expanding scope problem, is necessary right now, as MSBuildLocator current is broken for consumers that have VS2019 v16.7 installed.

If you want to solve the bigger problem, by all means do so - we will need to tackle it eventually - but for now we need something that gets users back up and running, and this PR does exactly that.

mattmurp:
> For completeness, the workaround is to copy the relevant `<dependentAssembly>` tags from MSBuild.exe.config to the config file of the application hosting MSBuild. It has the same downsides as this fix, i.e. it is not future-proof.

The binding redirects will work even if you don't ship the MSBuild assemblies, but only if you know what version of MSBuild you'll be using at runtime, which defeats the purpose of using MSBuildLocator.  So I would argue that the binding redirects solution has more downsides than the solution proposed in this PR.  While the solution in this PR may not be future-proof, it's the best solution currently available.

jzabroski:
> Wondering what others think, thanks!

I think modularity in .NET is a dumpster fire.  It doesn't stop with MSBuild.  The whole idea that this is an MSBuild issue is a joke.  Why is MSBuild solving modularity problems?

ladipro:
> The binding redirects will work even if you don't ship the MSBuild assemblies, but only if you know what version of MSBuild you'll be using at runtime, which defeats the purpose of using MSBuildLocator. So I would argue that the binding redirects solution has more downsides than the solution proposed in this PR. While the solution in this PR may not be future-proof, it's the best solution currently available.

That's a good point. I agree with this assessment and am in favor of taking this fix for now.

rainersigwald (approved):
This looks like a good option for now. There are other possibilities that might help more but this is better than the current state, so let's do it for now.

yaakov-h:
Thanks @rainersigwald!

Could you please publish a new release with this change to NuGet?

powercode:
Having `System.Runtime.CompilerServices.Unsafe` and `System.Numerics.Vector` in the array makes the locator say that MSBuild assemblies are already loaded if any of these two, completely unrelated, assemblies are loaded in the process.

I'd rather not see a build with this change.

rainersigwald:
@powercode that's a great point now tracked by #102.

PR #106, opened 2020-12-07
Only consider sdks from before the runtime

### LINKED ISSUE #92: Loading net5.0 Assemblies Shouldn't Happen When Running < net5.0
Related issue and repro: https://github.com/dotnet/msbuild/issues/5600

When targeting netcore3.1 and loading an assembly that depends on a net5.0 assembly (in this case `Microsoft.Build.NuGetSdkResolver` depending on `System.Runtime` version 5.0), it can fail to load.

It fails on line 118 of SdkResolverService
```
                 result = (SdkResult)sdkResolver.Resolve(sdk, context, resultFactory);
```
```
"System.IO.FileNotFoundException: Could not load file or assembly 'System.Runtime, Version=5.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'. The system cannot find the file specified.
File name: 'System.Runtime, Version=5.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'
   at Microsoft.Build.NuGetSdkResolver.NuGetSdkResolver.Resolve(SdkReference sdkReference, SdkResolverContext resolverContext, SdkResultFactory factory)
   at Microsoft.Build.BackEnd.SdkResolution.SdkResolverService.ResolveSdk(Int32 submissionId, SdkReference sdk, LoggingContext loggingContext, ElementLocation sdkReferenceLocation, String solutionPath, String projectPath, Boolean interactive)

"
```

We should consider changing MSBuildLocator to NOT return .NET Core SDK's that are newer than the current .NET Core runtime.

Forgind:
Fixes #92

This is now based on #79.


src/MSBuildLocator/DotNetSdkLocationHelper.cs
```diff
@@ -36,25 +35,15 @@ public static VisualStudioInstance GetInstance(string workingDirectory)
...
-
-            if (!versionMatch.Success)
-            {
-                return null;
-            }
-
-            if (!int.TryParse(versionMatch.Groups[1].Value, out int major) ||
-                !int.TryParse(versionMatch.Groups[2].Value, out int minor) ||
-                !int.TryParse(versionMatch.Groups[3].Value, out int patch))
+            if (!Version.TryParse(File.ReadAllText(versionPath), out Version version) || version > Environment.Version)
```
rainersigwald, line 38: `Environment.Version` is a strange beast. It's **always** `4.0.30319.42000` for .NET 4.6 and higher and for .NET Core 1.0-3.1. Starting with .NET 5.0, it returns a value more like you'd probably expect: https://github.com/dotnet/coreclr/pull/22664.

FORTUNATELY, since this PR is mostly about blocking out 5.0 versions from 3.1 hosts (since people mostly didn't complain about 3.1 versions for 2.1 hosts) this simple call should still work. And keep working into the future for 6.0/7.0 on 5.0 and so on.

src/MSBuildLocator/DotNetSdkLocationHelper.cs
```diff
@@ -14,7 +14,6 @@ namespace Microsoft.Build.Locator
     internal static class DotNetSdkLocationHelper
     {
         private static readonly Regex DotNetBasePathRegex = new Regex("Base Path:(.*)$", RegexOptions.Multiline);
-        private static readonly Regex VersionRegex = new Regex(@"^(\d+)\.(\d+)\.(\d+)", RegexOptions.Multiline);
```
rainersigwald, line 17: Have you tried your new code against a preview SDK? I think the additional complexity here over just using `Version.TryParse` is to handle a semver version like `6.0.100-alpha.234622`.

src/MSBuildLocator/DotNetSdkLocationHelper.cs
```diff
@@ -36,25 +35,19 @@ public static VisualStudioInstance GetInstance(string workingDirectory)
...
-            }
-
-            if (!int.TryParse(versionMatch.Groups[1].Value, out int major) ||
-                !int.TryParse(versionMatch.Groups[2].Value, out int minor) ||
-                !int.TryParse(versionMatch.Groups[3].Value, out int patch))
+            // Preview versions contain a hyphen after the numeric part of the version. Version.TryParse doesn't accept that.
+            string parseableVersion = File.ReadAllText(versionPath);
+            int indexOfHyphen = parseableVersion.IndexOf('-');
+            parseableVersion = indexOfHyphen >= 0 ? parseableVersion.Substring(0, indexOfHyphen) : parseableVersion;
+            if (!Version.TryParse(parseableVersion, out Version version) || version > Environment.Version)
```
ladipro, line 42: Should we first merge #79 and then move the version filtering to `RegisterDefaults`? I.e. the **default** behavior would be to pick the latest SDK not newer than the currently running runtime but the developer can implement their own policy if they choose so? There are legitimate use cases where the developer wants to know about all instances even if they are not compatible with the current process -- they might want to launch MSBuild out-of-proc for example.

rainersigwald: I'd like to get a minimal 5.0 fix out ASAP but I think this is a good way to go long-term.

Forgind: I made a change that I think implements this. I tested it with 5.0 and 3.1 runtimes, and they correctly identified 5.0 and 3.1 as the SDKs to use respectively, both ignoring my 2.2 SDK.

src/MSBuildLocator/MSBuildLocator.cs
```diff
@@ -93,7 +93,11 @@ public static IEnumerable<VisualStudioInstance> QueryVisualStudioInstances()
         /// <returns>Instance of Visual Studio found and registered.</returns>
         public static VisualStudioInstance RegisterDefaults()
         {
-            var instance = GetInstances(VisualStudioInstanceQueryOptions.Default).FirstOrDefault();
+            IEnumerable<VisualStudioInstance> instances = GetInstances(VisualStudioInstanceQueryOptions.Default);
+            VisualStudioInstance instance = instances.Where(
+                inst => inst.Version.Major < Environment.Version.Major || (inst.Version.Major == Environment.Version.Major && inst.Version.Minor <= Environment.Version.Minor)
+                ).FirstOrDefault() ??
+                instances.FirstOrDefault();
```
Forgind, line 100: Do you think I should keep this as a fallback? It only ever would return an SDK from after the runtime, but "No instances of MSBuild could be detected" would be misleading.

rainersigwald (changes requested):

src/MSBuildLocator/MSBuildLocator.cs
```diff
@@ -93,7 +93,11 @@ public static IEnumerable<VisualStudioInstance> QueryVisualStudioInstances()
         /// <returns>Instance of Visual Studio found and registered.</returns>
         public static VisualStudioInstance RegisterDefaults()
```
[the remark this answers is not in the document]
Forgind: I was trying to implement https://github.com/microsoft/MSBuildLocator/pull/106#discussion_r542216240

Is that wrong?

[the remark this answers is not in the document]
rainersigwald: Ah. Yes, I now think it's not the right design. We could potentially add another overload that takes a bool to "give me everything, I don't care" but I don't think we should filter in RegisterDefaults and I do think we should present only usable instances in the list by default.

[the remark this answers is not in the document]
Forgind: Reverted to filtering in DotNetSdkLocationHelper.cs.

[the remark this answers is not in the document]
Forgind: I'm a little resistant to add an overload for "give me everything" without anyone explicitly asking for it. (I'm guessing most people who use MSBuildLocator use it for relatively straightforward purposes.) It would also make it harder to pull in my unit testing PR when that comes along.

src/MSBuildLocator/DotNetSdkLocationHelper.cs
```diff
@@ -35,17 +35,15 @@ public static VisualStudioInstance GetInstance(string dotNetSdkPath)
...
+            // Preview versions contain a hyphen after the numeric part of the version. Version.TryParse doesn't accept that.
+            Match versionMatch = VersionRegex.Match(File.ReadAllText(versionPath));
 
-            if (!versionMatch.Success)
-            {
-                return null;
-            }
-
-            if (!int.TryParse(versionMatch.Groups[1].Value, out int major) ||
+            if (!versionMatch.Success ||
```
rainersigwald, line 41: I prefer the original flow here because the failure conditions are much more obvious in separate `if`s:

1. The version regex didn't match.
2. We can't parse a version out.
3. the version doesn't match

Here all together it's hard to see that there are separate failure cases.

src/MSBuildLocator/DotNetSdkLocationHelper.cs
```diff
@@ -35,17 +35,15 @@ public static VisualStudioInstance GetInstance(string dotNetSdkPath)
...
+                !int.TryParse(versionMatch.Groups[1].Value, out int major) ||
                 !int.TryParse(versionMatch.Groups[2].Value, out int minor) ||
-                !int.TryParse(versionMatch.Groups[3].Value, out int patch))
+                !int.TryParse(versionMatch.Groups[3].Value, out int patch) ||
+                major > Environment.Version.Major ||
+                (major == Environment.Version.Major && minor > Environment.Version.Minor))
```
rainersigwald, line 46: This definitely needs a comment! Describe all the things you've learned about why we care about the current runtime version relative to the SDK version, and the SDK policy as we currently understand it.

src/MSBuildLocator/MSBuildLocator.cs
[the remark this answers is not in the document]
rainersigwald: How would this affect a unit testing PR?

rainersigwald (changes requested):

src/MSBuildLocator/DotNetSdkLocationHelper.cs
```diff
@@ -49,6 +49,16 @@ public static VisualStudioInstance GetInstance(string dotNetSdkPath)
             {
                 return null;
             }
+            
+            // Components of the SDK often have dependencies on the runtime they shipped with, including that several tasks that shipped
+            // in the .NET 5 SDK rely on the .NET 5.0 runtime. Assuming the runtime that shipped with a particular SDK has the same version,
+            // this ensures that we don't choose an SDK that doesn't work with the chosen runtime. This is not guaranteed to always work but
```
rainersigwald, line 55:
```suggestion
            // this ensures that we don't choose an SDK that doesn't work with the runtime of the current application. This is not guaranteed to always work but
```

?

rainersigwald (approved):
LGTM. Let's get it out!

src/MSBuildLocator/MSBuildLocator.cs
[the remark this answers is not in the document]
Forgind: Two reasons: it would add something else I'd probably want to test, and since I'm already planning to add an overload to this, I'd have to split my overload into "give me everything" and "don't give me everything" cases. Not a huge deal, though.

PR #107, opened 2020-12-08
Load all assemblies from the MSBuild folder if not already loaded

### LINKED ISSUE #38: Should System.Collections.Immutable and System.Threading.Tasks.Dataflow be supported?
The Microsoft.Build assembly references the System.Collections.Immutable and System.Threading.Tasks.Dataflow assemblies, both living in the MSBuild build folder. I saw failures building WPF projects in our build application and I got around it by adding assembly resolving support for System.Collections.Immutable.

### LINKED ISSUE #103: .net 5 works poorly with latest preview prerelase.
When using a XUnit test project, `System.Numerics.Vectors` and `System.Runtime.CompilerServices.Unsafe` are already loaded in the process when the a module initializer is run. That is as early as can be.

```cs
	public static class ModInit
	{
		[ModuleInitializer]
		public static void Init()
		{
			try
			{				
				if (!MSBuildLocator.IsRegistered)
				{
					MSBuildLocator.RegisterDefaults();
				}
			}
			catch (InvalidOperationException ex)
			{
				System.Diagnostics.Trace.WriteLine(ex);
			}
		}
	}
```

```text
Microsoft.Build.Locator.MSBuildLocator.RegisterInstance was called, but MSBuild assemblies were already loaded.
Ensure that RegisterInstance is called before any method that directly references types in the Microsoft.Build namespace has been called.
This dependency arises from when a method is just-in-time compiled, so if it breaks even if the reference to a Microsoft.Build type has not been executed.
For more details, see aka.ms/RegisterMSBuildLocator
Loaded MSBuild assemblies: System.Numerics.Vectors, Version=5.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a
System.Runtime.CompilerServices.Unsafe, Version=5.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a
```

So having these assemblies in the vector

```cs
        private static readonly string[] s_msBuildAssemblies =
        {
            "Microsoft.Build",
            "Microsoft.Build.Engine",
            "Microsoft.Build.Framework",
            "Microsoft.Build.Tasks.Core",
            "Microsoft.Build.Utilities.Core",
            "System.Runtime.CompilerServices.Unsafe",
            "System.Numerics.Vectors"
        };
```
seems like a bad idea.

If they are needed as a dependency, that should be separated from if there are msbuild assemblies loaded in the process.

Forgind:
Changes the check for previously loaded assemblies to only fail if we already registered them via MSBuildLocator. Then adds previously loaded assemblies to the list of loaded assemblies so they aren't loaded again.

This may cause version mismatch issues, but if you had previously loaded some but not all of our assemblies, I'd say that's something for you to figure out. 😄 

Fixes #103.

Also Fixes #38.

### DIFF
--- a/samples/BuilderApp/BuilderApp.csproj
+++ b/samples/BuilderApp/BuilderApp.csproj
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFrameworks>net471;netcoreapp3.1</TargetFrameworks>
+    <TargetFrameworks>net471;netcoreapp2.1</TargetFrameworks>
     <CopyLocalLockFileAssemblies>false</CopyLocalLockFileAssemblies>
     <IsPackable>false</IsPackable>
   </PropertyGroup>
--- a/samples/BuilderApp/BuilderApp.csproj
+++ b/samples/BuilderApp/BuilderApp.csproj
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFrameworks>net471;netcoreapp2.1</TargetFrameworks>
+    <TargetFrameworks>net471;netcoreapp3.1</TargetFrameworks>
     <CopyLocalLockFileAssemblies>false</CopyLocalLockFileAssemblies>
     <IsPackable>false</IsPackable>
   </PropertyGroup>
--- a/src/MSBuildLocator/MSBuildLocator.cs
+++ b/src/MSBuildLocator/MSBuildLocator.cs
@@ -206,21 +206,6 @@ namespace Microsoft.Build.Locator
                     string targetAssembly = Path.Combine(msbuildPath, assemblyName.Name + ".dll");
                     if (File.Exists(targetAssembly))
                     {
-                        // If the assembly was already loaded, return that. This could theoretically cause version problems, but the user shouldn't be
-                        // trying to load an MSBuild that is incompatible with the version of the assembly they currently have loaded.
-                        Assembly loadedAssembly = AppDomain.CurrentDomain.GetAssemblies().FirstOrDefault(a => a.GetName().Name.Equals(assemblyName.Name));
-                        if (loadedAssembly != null)
-                        {
-                            loadedAssemblies.Add(assemblyName.FullName, loadedAssembly);
-                            return loadedAssembly;
-                        }
-
-                        // Automatically unregister the handler once all supported assemblies have been loaded.
-                        if (Interlocked.Increment(ref numResolvedAssemblies) == s_msBuildAssemblies.Length)
-                        {
-                            Unregister();
-                        }
-
                         assembly = Assembly.LoadFrom(targetAssembly);
                         loadedAssemblies.Add(assemblyName.FullName, assembly);
                         return assembly;
--- a/src/MSBuildLocator/MSBuildLocator.cs
+++ b/src/MSBuildLocator/MSBuildLocator.cs
@@ -35,7 +35,7 @@ namespace Microsoft.Build.Locator
 #endif
 
         // Used to determine when it's time to unregister the registeredHandler.
-        private static int numResolvedAssemblies;
+        private static bool registerCalled = false;
 
         /// <summary>
         ///     Gets a value indicating whether an instance of MSBuild is currently registered.
@@ -142,6 +142,7 @@ namespace Microsoft.Build.Locator
         /// </param>
         public static void RegisterMSBuildPath(string msbuildPath)
         {
+            registerCalled = true;
             if (string.IsNullOrWhiteSpace(msbuildPath))
             {
                 throw new ArgumentException("Value may not be null or whitespace", nameof(msbuildPath));
@@ -227,7 +228,7 @@ namespace Microsoft.Build.Locator
             if (!IsRegistered)
             {
                 var error = $"{typeof(MSBuildLocator)}.{nameof(Unregister)} was called, but no MSBuild instance is registered." + Environment.NewLine;
-                if (numResolvedAssemblies == 0)
+                if (!registerCalled)
                 {
                     error += $"Ensure that {nameof(RegisterInstance)}, {nameof(RegisterMSBuildPath)}, or {nameof(RegisterDefaults)} is called before calling this method.";
                 }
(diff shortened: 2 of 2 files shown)


src/MSBuildLocator/MSBuildLocator.cs
```diff
@@ -200,27 +198,32 @@ Assembly TryLoadAssembly(AssemblyName assemblyName)
...
-                                Unregister();
-                            }
-
-                            assembly = Assembly.LoadFrom(targetAssembly);
-                            loadedAssemblies.Add(assemblyName.FullName, assembly);
-                            return assembly;
+                            loadedAssemblies.Add(assemblyName.FullName, loadedAssembly);
+                            return loadedAssembly;
                         }
+
```
Forgind, line 217: delete

samples/BuilderApp/BuilderApp.csproj
```diff
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFrameworks>net471;netcoreapp2.1</TargetFrameworks>
+    <TargetFrameworks>net471;netcoreapp3.1</TargetFrameworks>
```
rainersigwald, line 5: Is this needed?

src/MSBuildLocator/MSBuildLocator.cs
```diff
@@ -37,7 +35,7 @@ public static class MSBuildLocator
 #endif
 
         // Used to determine when it's time to unregister the registeredHandler.
```
rainersigwald, line 37: Stale comment

src/MSBuildLocator/MSBuildLocator.cs
```diff
@@ -200,27 +199,17 @@ Assembly TryLoadAssembly(AssemblyName assemblyName)
...
                 lock (loadedAssemblies)
                 {
-                    Assembly assembly;
-                    if (loadedAssemblies.TryGetValue(assemblyName.FullName, out assembly))
+                    if (loadedAssemblies.TryGetValue(assemblyName.FullName, out Assembly assembly))
                     {
                         return assembly;
                     }
 
-                    if (IsMSBuildAssembly(assemblyName))
```
rainersigwald, line 209: Delete no-longer-needed `IsMSBuildAssembly` method. It's used in `LoadedMsBuildAssemblies` but I think that should now just return what's currently the local value `loadedAssemblies` (I guess it needs to become instance).

src/MSBuildLocator/MSBuildLocator.cs
```diff
@@ -200,27 +199,17 @@ Assembly TryLoadAssembly(AssemblyName assemblyName)
...
+                    string targetAssembly = Path.Combine(msbuildPath, assemblyName.Name + ".dll");
```
rainersigwald, line 207:
```suggestion
                    // Look in the MSBuild folder for any unresolved reference--it may be a dependency
                    // of MSBuild or a task.
                    string targetAssembly = Path.Combine(msbuildPath, assemblyName.Name + ".dll");
```

src/MSBuildLocator/MSBuildLocator.cs
```diff
@@ -239,7 +228,7 @@ public static void Unregister()
             if (!IsRegistered)
             {
                 var error = $"{typeof(MSBuildLocator)}.{nameof(Unregister)} was called, but no MSBuild instance is registered." + Environment.NewLine;
-                if (numResolvedAssemblies == 0)
+                if (!registerCalled)
                 {
                     error += $"Ensure that {nameof(RegisterInstance)}, {nameof(RegisterMSBuildPath)}, or {nameof(RegisterDefaults)} is called before calling this method.";
```
rainersigwald, line 233: I think this whole error system might be kinda useless now--we no longer really have the concept of `Unregister` since the assemblies we might resolve is now unbounded. This change is OK but I think it's worth thinking of just rewriting this whole thing to do nothing.

rainersigwald (changes requested):

src/MSBuildLocator/MSBuildLocator.cs
```diff
@@ -50,9 +48,7 @@ public static class MSBuildLocator
         /// <remarks>
         ///     If any Microsoft.Build assemblies are already loaded into the current AppDomain, the value will be false.
         /// </remarks>
-        public static bool CanRegister => !IsRegistered && !LoadedMsBuildAssemblies.Any();
-
-        private static IEnumerable<Assembly> LoadedMsBuildAssemblies => AppDomain.CurrentDomain.GetAssemblies().Where(IsMSBuildAssembly);
+        public static bool CanRegister => !IsRegistered && !loadedAssemblies.Any();
```
rainersigwald, line 51: Naturally looking at the diff that's the result of the change I asked you to make revealed to me that I was wrong and this isn't what we want. I'll push a fix.

src/MSBuildLocator/MSBuildLocator.cs
```diff
@@ -239,7 +230,7 @@ public static void Unregister()
             if (!IsRegistered)
             {
                 var error = $"{typeof(MSBuildLocator)}.{nameof(Unregister)} was called, but no MSBuild instance is registered." + Environment.NewLine;
-                if (numResolvedAssemblies == 0)
+                if (!registerCalled)
```
Forgind, line 233: This error message is now wrong. Unregister isn't called automatically, so the only way IsRegistered would be false but registerCalled would be true is if they manually Unregistered twice, which probably isn't a big thing. I'd prefer to unset registerCalled here and keep just the one message.

src/MSBuildLocator/MSBuildLocator.cs
```diff
@@ -37,7 +35,7 @@ public static class MSBuildLocator
 #endif
 
         // Used to determine when it's time to unregister the registeredHandler.
-        private static int numResolvedAssemblies;
+        private static bool registerCalled = false;
```
Forgind, line 38: Stale comment—we don't unregister the registeredHandler.

src/MSBuildLocator/MSBuildLocator.cs
rainersigwald: I'm not 100% sure I understand what you mean but go ahead and do that.

rainersigwald (approved):

PR #109, opened 2020-12-10
Indicates whether Roslyn is included

### LINKED ISSUE #22: Provide a way for consumers to validate a VS instance
OmniSharp ships a very similar MSBuild discovery mechanism to this one. However, it verifies that the VS instance that it considers has the "Microsoft.VisualStudio.Components.Roslyn.Compiler" package installed.

https://github.com/OmniSharp/omnisharp-roslyn/blob/master/src/OmniSharp.Host/MSBuild/Discovery/Providers/VisualStudioInstanceProvider.cs#L56-L57

It would be great to provide consumers with some way to validate an instance in a similar manner. Otherwise, it's difficult to determine whether an instance is actually a reasonable choice.

Forgind:
Fixes #22

I didn't see an obvious way to determine whether the C# compiler was installed for the dotnet sdk, and I didn't want to replicate all the logic from VSLocationHelper for a command prompt. Maybe right?

### DIFF
--- a/src/MSBuildLocator/DotNetSdkLocationHelper.cs
+++ b/src/MSBuildLocator/DotNetSdkLocationHelper.cs
@@ -55,7 +55,8 @@ namespace Microsoft.Build.Locator
                 name: ".NET Core SDK",
                 path: dotNetSdkPath,
                 version: new Version(major, minor, patch),
-                discoveryType: DiscoveryType.DotNetSdk);
+                discoveryType: DiscoveryType.DotNetSdk,
+                true);
         }
 
         private static string GetDotNetBasePath(string workingDirectory)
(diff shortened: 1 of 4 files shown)


rainersigwald (changes requested):
I don't think this is the right approach. Is it possible to expose `instance` itself so consumers can do arbitrary operations on it, like check for their own packages?

Forgind:
I could expose instance as a parameter as I have ContainsRoslynCompiler exposed here. Would you want the non-VS versions to return null for that?

YuliiaKovalova:
Hi @rainersigwald and @Forgind,

Do we plan to work on this PR?
Can we close it?

rainersigwald:
I think we need to have a design for #22 before taking any implementation and would favor closing this.

Forgind:
I think this is low enough priority that it doesn't warrant a serious design discussion. I'd personally prefer to just take this roughly as-is over closing it, but closing it makes more sense than talking about it for too long.

PR #110, opened 2020-12-11
Update TF versions

Forgind:


src/MSBuildLocator/Microsoft.Build.Locator.csproj
```diff
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <OutputType>Library</OutputType>
-    <TargetFrameworks>net46;netcoreapp2.0</TargetFrameworks>
+    <TargetFrameworks>net472;netcoreapp3.1</TargetFrameworks>
```
rainersigwald, line 5: how did you pick this set of TFs?

I don't think we should update the .NET Framework target version; we should match what we targeted for VS2017 so that MSBuild is usable in client applications.

For core I'm less sure. Should we target 2.1, the oldest currently-in-support runtime? Should we target net5.0, the newest? should we multitarget?

src/MSBuildLocator.Tests/Microsoft.Build.Locator.Tests.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>net46;netcoreapp2.0</TargetFrameworks>
+    <TargetFrameworks>net472;netcoreapp3.1</TargetFrameworks>
```
rainersigwald, line 4:
```suggestion
    <TargetFrameworks>net46;netcoreapp3.1;net5.0</TargetFrameworks>
```

samples/BuilderApp/BuilderApp.csproj
```diff
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFrameworks>net471;netcoreapp2.1</TargetFrameworks>
+    <TargetFrameworks>net472;netcoreapp3.1</TargetFrameworks>
```
rainersigwald, line 5: I'd do 5.0 here too

rainersigwald (changes requested):

src/MSBuildLocator/Microsoft.Build.Locator.csproj
Forgind: Choosing 5.0 would apparently make it easier, since 2.0 and 5.0 are installed but not 3.1...

Forgind: Personally, I think 3.1 makes sense as in-support, going to stay in support longer than 2.1, and still before net5.0, which I'm guessing is somewhat less commonly used at this point. I'm happy switching to 5.0, though. If we go with that, I assume that also means I should have parts that are just netcoreapp but make the rest of the code run all the time?

samples/BuilderApp/BuilderApp.csproj
```diff
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFrameworks>net472;netcoreapp2.1;net5.0</TargetFrameworks>
+    <TargetFrameworks>net46;netcoreapp3.1;net5.0</TargetFrameworks>
```
rainersigwald, line 5: No need to back these down IMO, we only really care about the main library's TFs.
```suggestion
    <TargetFrameworks>net472;netcoreapp3.1;net5.0</TargetFrameworks>
```

src/MSBuildLocator.Tests/Microsoft.Build.Locator.Tests.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>net472;netcoreapp2.1</TargetFrameworks>
+    <TargetFrameworks>net46;netcoreapp3.1</TargetFrameworks>
```
rainersigwald, line 4:
```suggestion
    <TargetFrameworks>net472;netcoreapp3.1</TargetFrameworks>
```

src/MSBuildLocator/Microsoft.Build.Locator.csproj
```diff
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <OutputType>Library</OutputType>
-    <TargetFrameworks>net46;netcoreapp2.1</TargetFrameworks>
+    <TargetFrameworks>net46;netcoreapp3.1</TargetFrameworks>
```
rainersigwald, line 5: I think this needs a version bump; it's technically semver major but I don't know if we need to go that far but it seems weird to do it in a patch-level change (default for no `version.json` changes).

samples/BuilderApp/BuilderApp.csproj
Forgind: I was thinking it seems weird to test something that's at all different from the real product. Don't think it matters too much, though.

.github/workflows/pull-request.yml
```diff
@@ -17,14 +17,14 @@ jobs:
...
+    - name: Setup .NET Core 3.1 runtime
       uses: actions/setup-dotnet@v1
       with:
-        dotnet-version: 2.1.x
+        dotnet-version: 3.1.x
     - name: Setup .NET SDK
       uses: actions/setup-dotnet@v1
       with:
-        dotnet-version: 5.0.x
+        dotnet-version: 6.0.x
```
rainersigwald, line 27: There's a simpler way now:

```yaml
    - name: Setup .NET SDK
      uses: actions/setup-dotnet@v1
      with:
        dotnet-version: |
          3.1.x
          6.0.x
```

rainersigwald (approved):

Forgind:
I'd prefer to merge the hostfxr change first.

PR #112, opened 2020-12-17
Add summary blurb to README

rainersigwald:
@marcpopMSFT, @forgind, @yevgeni-zolotko -- would having this text here have helped y'all understand the intention of MSBuildLocator and why it's necessary over just "MSBuild NuGet package references"?

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,13 @@
 # Microsoft.Build.Locator
 
+## What is Locator for?
+
+MSBuild offers a .NET API surface that allows you to [evaluate](https://docs.microsoft.com/dotnet/api/microsoft.build.evaluation) and [build](https://docs.microsoft.com/dotnet/api/microsoft.build.execution) MSBuild projects from an application. MSBuild is available as a set of NuGet packages that can provide the implementation of the MSBuild programming language to your application. But it's generally not enough to be able to load *some* MSBuild project: you want your application to be able to load the `.csproj`, `.vbproj`, `.fsproj`, `.sqlproj`, `.ccproj` and other project types that you can build in Visual Studio (or Visual Studio Build Tools) or with the .NET SDK. To do that, you need more than just MSBuild's assemblies--you must also have access to all of the SDKs and build logic that are imported into those projects.
+
+That additional build logic is distributed with Visual Studio, with Visual Studio extensions, or as part of the .NET SDK. So to correctly load projects, you need to load them in the context of one of those MSBuild installations.
+
+Loading MSBuild from Visual Studio also ensures that your application gets the same view of projects as `MSBuild.exe`, `dotnet build`, or Visual Studio, including bug fixes, feature additions, and performance improvements that may come from a newer MSBuild release.
+
 ## Documentation
 
 Documentation is located on the official Microsoft documentation site: [Use Microsoft.Build.Locator](https://docs.microsoft.com/visualstudio/msbuild/updating-an-existing-application#use-microsoftbuildlocator).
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,5 +1,13 @@
 # Microsoft.Build.Locator
 
+## What is Locator for?
+
+MSBuild offers a .NET API surface that allows you to [evaluate](https://docs.microsoft.com/dotnet/api/microsoft.build.evaluation) and [build](https://docs.microsoft.com/dotnet/api/microsoft.build.execution) MSBuild projects from an application. MSBuild is available as a set of NuGet packages that can provide the implementation of the MSBuild programming language to your application. But it's generally not enough to be able to load *some* MSBuild project: you want your application to be able to load the `.csproj`, `.vbproj`, `.fsproj`, `.sqlproj`, `.ccproj` and other project types that you can build in Visual Studio (or Visual Studio Build Tools) or with the .NET SDK. To do that, you need more than just MSBuild's assemblies--you must also have access to all of the SDKs and build logic that are imported into those projects.
```
Forgind, line 5:
```suggestion
MSBuild offers a .NET API surface that allows you to [evaluate](https://docs.microsoft.com/dotnet/api/microsoft.build.evaluation) and [build](https://docs.microsoft.com/dotnet/api/microsoft.build.execution) MSBuild projects from an application. MSBuild is available as a set of NuGet packages that can provide the implementation of the MSBuild programming language to your application. But it's generally not enough to be able to load *some* MSBuild project: you want your application to be able to load the `.*proj` that you can build in Visual Studio (or Visual Studio Build Tools) or with the .NET SDK. To do that, you need more than just MSBuild's assemblies--you must also have access to all of the SDKs and build logic that are imported into those projects.
```

Without other information, I would assume that `it's generally not enough to be able to load *some* MSBuild project` means that you can load any MSBuild project that doesn't have too much custom logic. Can you specify what you mean by build logic? My mind would go to "code in another assembly" before ".props/.targets", and I'd wonder why you can't pull that from a NuGet package, too. Also, since I got stuck for quite a while on "why can't I just call the API (from a nupkg) and let you figure out the implementation at runtime", that might be helpful to explain here. New .NET developers might not know that there are multiple SDKs (I didn't realize that for my first few months), which might be confusing.

rainersigwald: I prefer the explicit list because it helps convey that there are different types of project, not just "msbuild projects".

> Can you specify what you mean by build logic?

I mean MSBuild XML files (and the tasks they call).

> Also, since I got stuck for quite a while on "why can't I just call the API (from a nupkg) and let you figure out the implementation at runtime", that might be helpful to explain here.

I don't understand this comment. How would you explain that?

> New .NET developers might not know that there are multiple SDKs (I didn't realize that for my first few months), which might be confusing.

What do you mean by "multiple SDKs" here? There are SDKs for all kinds of stuff--Windows, Azure, Python . . .

Forgind: > I mean MSBuild XML files (and the tasks they call).

Can you include that in the readme?

> I don't understand this comment. How would you explain that?

It's related to the build logic problem. Perhaps "Some of the properties, tasks, targets, and so on included in a standard VS installation, VS Build Tools, and the .NET SDK are proprietary, including some you couldn't find in NuGet packages. This means that although you could reference parts of the .NET API surface that purport to load a project, evaluate it, or build it, doing so would lead to a failure at runtime.

Some of that could probably be integrated into the next paragraph.

> What do you mean by "multiple SDKs" here? There are SDKs for all kinds of stuff--Windows, Azure, Python . . .

Exactly! I didn't realize that for quite a while. "The .NET SDK" sounds like it is the _only_ SDK relevant to .NET developers. As someone who had never programmed with .NET or C# or even (as far as I can remember) used anything that called itself an SDK, I didn't know there were others, and I'd imagine either that that was a typo, or the .NET SDK somehow had the .NET SDK inside, which wouldn't make sense.

AndyGerlicher (approved):

PR #115, opened 2020-12-18
Load NuGet assemblies

### LINKED ISSUE #86: Should MSBuildLocator handle loading of NuGet assemblies?
Using this with .NET sdk 3.1.X - I get errors like the following if I run any build targets.

```txt
Build FAILED.

/usr/local/share/dotnet/sdk/3.1.102/Sdks/Microsoft.NET.Sdk/targets/Microsoft.NET.Sdk.FrameworkReferenceResolution.targets(59,5): error MSB4018: The "ProcessFrameworkReferences" task failed unexpectedly.
/usr/local/share/dotnet/sdk/3.1.102/Sdks/Microsoft.NET.Sdk/targets/Microsoft.NET.Sdk.FrameworkReferenceResolution.targets(59,5): error MSB4018: System.IO.FileNotFoundException: Could not load file or assembly 'NuGet.Frameworks, Version=5.4.0.2, Culture=neutral, PublicKeyToken=31bf3856ad364e35'. The system cannot find the file specified.
/usr/local/share/dotnet/sdk/3.1.102/Sdks/Microsoft.NET.Sdk/targets/Microsoft.NET.Sdk.FrameworkReferenceResolution.targets(59,5): error MSB4018:
/usr/local/share/dotnet/sdk/3.1.102/Sdks/Microsoft.NET.Sdk/targets/Microsoft.NET.Sdk.FrameworkReferenceResolution.targets(59,5): error MSB4018: File name: 'NuGet.Frameworks, Version=5.4.0.2, Culture=neutral, PublicKeyToken=31bf3856ad364e35'
/usr/local/share/dotnet/sdk/3.1.102/Sdks/Microsoft.NET.Sdk/targets/Microsoft.NET.Sdk.FrameworkReferenceResolution.targets(59,5): error MSB4018:    at System.Linq.Enumerable.WhereEnumerableIterator`1.ToList()
/usr/local/share/dotnet/sdk/3.1.102/Sdks/Microsoft.NET.Sdk/targets/Microsoft.NET.Sdk.FrameworkReferenceResolution.targets(59,5): error MSB4018:    at System.Linq.Enumerable.ToList[TSource](IEnumerable`1 source)
/usr/local/share/dotnet/sdk/3.1.102/Sdks/Microsoft.NET.Sdk/targets/Microsoft.NET.Sdk.FrameworkReferenceResolution.targets(59,5): error MSB4018:    at Microsoft.NET.Build.Tasks.ProcessFrameworkReferences.ExecuteCore()
/usr/local/share/dotnet/sdk/3.1.102/Sdks/Microsoft.NET.Sdk/targets/Microsoft.NET.Sdk.FrameworkReferenceResolution.targets(59,5): error MSB4018:    at Microsoft.NET.Build.Tasks.TaskBase.Execute()
/usr/local/share/dotnet/sdk/3.1.102/Sdks/Microsoft.NET.Sdk/targets/Microsoft.NET.Sdk.FrameworkReferenceResolution.targets(59,5): error MSB4018:    at Microsoft.Build.BackEnd.TaskExecutionHost.Microsoft.Build.BackEnd.ITaskExecutionHost.Execute()
/usr/local/share/dotnet/sdk/3.1.102/Sdks/Microsoft.NET.Sdk/targets/Microsoft.NET.Sdk.FrameworkReferenceResolution.targets(59,5): error MSB4018:    at Microsoft.Build.BackEnd.TaskBuilder.ExecuteInstantiatedTask(ITaskExecutionHost taskExecutionHost, TaskLoggingContext taskLoggingContext, TaskHost taskHost, ItemBucket bucket, TaskExecutionMode howToExecuteTask)
/usr/local/share/dotnet/sdk/3.1.102/Sdks/Microsoft.NET.Sdk/targets/Microsoft.NET.Sdk.FrameworkReferenceResolution.targets(59,5): error MSB4018:
/usr/local/share/dotnet/sdk/3.1.102/Sdks/Microsoft.NET.Sdk/targets/Microsoft.NET.Sdk.FrameworkReferenceResolution.targets(59,5): error MSB4018:
    0 Warning(s)
    1 Error(s)
```

I think that's happening here is that the tasks assembly `Microsoft.NET.Build.Tasks.dll` has references to NuGet's assemblies. If this were running as part of a dedicated msbuild process it would be fine because these would be in the load context. However since I'm hosting msbuild in my process they are not. 

I was able to work around this with my own assembly resolve handler, but it doesn't feel good to maintain a lot of assemblies that msbuild is usually responsible for providing.

Forgind:
This is based off of #107. Review that first.

That PR means we won't error if you load an assembly that we also know how to load. This PR teaches MSBuildLocator how to find NuGet assemblies in VS scenarios.

Fixes #86.


src/MSBuildLocator/MSBuildLocator.cs
```diff
@@ -210,6 +210,15 @@ Assembly TryLoadAssembly(AssemblyName assemblyName)
                         return assembly;
                     }
 
+                    // Finds and loads NuGet assemblies if msbuildPath is in a VS installation
```
rainersigwald, line 213: should this be wrapped in an `#if NETFRAMEWORK` or something so we don't waste the time on Core?

src/MSBuildLocator/MSBuildLocator.cs
```diff
@@ -210,6 +210,15 @@ Assembly TryLoadAssembly(AssemblyName assemblyName)
...
+                    targetAssembly = Path.GetFullPath(Path.Combine(msbuildPath, "..", "..", "..", "Common7", "IDE", "CommonExtensions", "Microsoft", "NuGet", assemblyName.Name + ".dll"));
```
rainersigwald, line 214: Instead of hardcoding this relative path, should we add an argument 

```C#
public static void RegisterMSBuildPath(string msbuildPath, string[] additionalDirectories)
```

And use that in `RegisterInstance`?

src/MSBuildLocator/MSBuildLocator.cs
```diff
@@ -139,14 +139,40 @@ public static void RegisterInstance(VisualStudioInstance instance)
         /// </param>
         public static void RegisterMSBuildPath(string msbuildPath)
         {
-            if (string.IsNullOrWhiteSpace(msbuildPath))
+            RegisterMSBuildPath(new string[] {
+                msbuildPath
+#if NET46
```
rainersigwald, line 144: I think this needs an if-not-mono, right?

src/MSBuildLocator/MSBuildLocator.cs
```diff
@@ -139,14 +139,40 @@ public static void RegisterInstance(VisualStudioInstance instance)
...
+        public static void RegisterMSBuildPath(string[] msbuildSearchPaths)
+        {
+            if (msbuildSearchPaths.Length < 1)
+            {
+                throw new ArgumentException("Must provide at least one search path to RegisterMSBuildPath.");
+            }
+            if (msbuildSearchPaths.Any(string.IsNullOrWhiteSpace))
             {
-                throw new ArgumentException("Value may not be null or whitespace", nameof(msbuildPath));
+                throw new ArgumentException("Value may not be null or whitespace", nameof(msbuildSearchPaths));
```
rainersigwald, line 170: Include index of bogus entry?

src/MSBuildLocator/MSBuildLocator.cs
```diff
@@ -202,21 +228,15 @@ Assembly TryLoadAssembly(AssemblyName assemblyName)
...
-                    targetAssembly = Path.GetFullPath(Path.Combine(msbuildPath, "..", "..", "..", "Common7", "IDE", "CommonExtensions", "Microsoft", "NuGet", assemblyName.Name + ".dll"));
-                    if (File.Exists(targetAssembly))
-                    {
-                        assembly = Assembly.LoadFrom(targetAssembly);
-                        loadedAssemblies.Add(assemblyName.FullName, assembly);
-                        return assembly;
+                        string targetAssembly = Path.Combine(msbuildPath, assemblyName.Name + ".dll");
+                        if (File.Exists(targetAssembly))
+                        {
+                            assembly = Assembly.LoadFrom(targetAssembly);
```
rainersigwald, line 236: Not unique to this PR but should we improve error handling here? Like if LoadFrom fails because it exists but is a native DLL or something? Shouldn't be a major problem since the naming we're searching is very restricted.

src/MSBuildLocator/MSBuildLocator.cs
Forgind: I agree, but it's worth noting that without it, it would just do an extra check without actually failing.

Forgind: Also occurred to me that we don't support mono quite yet, but that's changing soon.

src/MSBuildLocator/MSBuildLocator.cs
Forgind: 0-indexed or 1-indexed? I made it 1-indexed but would be happy to change it.

src/MSBuildLocator/MSBuildLocator.cs
Forgind: I haven't checked to see what (or how bad) the error message is for an invalid dll, but it would seem unnecessarily confusing to put out two assemblies with identical names and extensions in (at least roughly) the same spot.

rainersigwald: Yeah, let's just leave it.

src/MSBuildLocator/MSBuildLocator.cs
```diff
@@ -139,14 +139,50 @@ public static void RegisterInstance(VisualStudioInstance instance)
         /// </param>
         public static void RegisterMSBuildPath(string msbuildPath)
         {
-            if (string.IsNullOrWhiteSpace(msbuildPath))
+            RegisterMSBuildPath(new string[] {
+                msbuildPath
+#if NET46 && !MONO
```
rainersigwald, line 144: Because mono runs "unmodified" .NET Framework binaries, it doesn't usually get a compile-time constant (and can't easily have a specific binary in a NuGet package). I think this check would need to happen at runtime and use `IsRunningOnMono` introduced in #82. Since that isn't landed yet, let's just drop it here and add a note to that PR to merge this and add the check?

rainersigwald (approved):

src/MSBuildLocator/MSBuildLocator.cs
Forgind: Sounds good to me.

Forgind, line 144:
```suggestion
#if NET46
```

PR #116, opened 2020-12-24
Analyzer

### LINKED ISSUE #67: Roslyn Analyzer
There's  a very common problem using Locator, which is that the natural code change to use it:

```diff
 void DoStuff()
 {
+    MSBuildLocator.RegisterDefaults();
     var project = new Microsoft.Build.Evaluation.Project(projectPath);
```

Doesn't work, because .NET loads all types referenced in a method before executing that method, so it tries to load MSBuild APIs before the Locator has made that possible.

#64 is one example of this but there have been a ton.

We can't fix this in the deployed code, because it won't have been loaded in time to do anything.

It might be possible, though, to create a Roslyn analyzer that ships as part of this package that checks to make sure there are no MSBuild APIs in a method that calls the locator. That would make things much easier for new users.

Forgind:
image
vs.
image
image
image
image

Still to do: drop into vsix.

Fixes #67


src/MSBuildLocatorAnalyzer/NuGet.config
```diff
@@ -0,0 +1,11 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<configuration>
+  <packageSources>
+    <clear />
+    <add key="nuget.org" value="https://api.nuget.org/v3/index.json" />
+    <add key="dotnet-tools" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-tools/nuget/v3/index.json" />
+  </packageSources>
```
Youssef1313, line 7: ⚠️ ⚠️ These feeds shouldn't be mixed.

src/MSBuildLocatorAnalyzer/MSBuildLocatorAnalyzerAnalyzer.cs
```diff
@@ -0,0 +1,75 @@
...
+        {
+            context.RegisterOperationAction(opact =>
+            {
+                var invocation = (IInvocationOperation)opact.Operation;
+                var targetMethod = invocation.TargetMethod;
+                if (targetMethod?.ContainingAssembly != null)
+                {
+                    if (targetMethod.ContainingAssembly.Name.Equals("Microsoft.Build.Locator") && msbuildLocatorRegisterMethods.Contains(targetMethod.Name))
+                    {
+                        MethodDeclarationSyntax method = (MethodDeclarationSyntax)invocation.Syntax.FirstAncestorOrSelf((Func<SyntaxNode, bool>)(a => a is MethodDeclarationSyntax));
```
Youssef1313, line 55: Why do this? I don't think syntax is any important here. e.g, this will be a false negative in a local function or constructor for example. But I'm not suggesting to check against more syntax kinds, instead, I suggest removing the `method != null` check completely and remove this variable as well.

src/MSBuildLocatorAnalyzer/MSBuildLocatorAnalyzerAnalyzer.cs
```diff
@@ -0,0 +1,75 @@
...
+                            for (symbolParent = invocation; symbolParent.Parent != null && symbolParent.Kind != OperationKind.MethodBody; symbolParent = symbolParent.Parent);
+                            foreach (var child in symbolParent.Descendants())
+                            {
+                                if ((child is IInvocationOperation op && msBuildAssemblies.Contains(op.TargetMethod?.ContainingAssembly?.Name)) ||
+                                (child is ITypeOfOperation toOp && msBuildAssemblies.Contains(toOp.TypeOperand.ContainingAssembly?.Name)) ||
+                                (msBuildAssemblies.Contains(child.Type?.ContainingAssembly?.Name)))
+                                {
+                                    opact.ReportDiagnostic(Diagnostic.Create(Rule, child.Syntax.GetLocation()));
+                                }
+                            }
```
Youssef1313, line 68: A different approach would be something like this:

```csharp
context.RegisterOperationBlockStartAction(context =>
{
    if (context.OwningSymbol.Kind != SymbolKind.Method)
    {
        return;
    }

    var hasRegisterMethod = false;
    var hasBadReference = false;
    
    context.RegisterOperationAction(context =>
    {
        var invocation = (IInvocationOperation)opact.Operation;
        var targetMethod = invocation.TargetMethod;
        if (targetMethod.ContainingAssembly.Name.Equals("Microsoft.Build.Locator") && msbuildLocatorRegisterMethods.Contains(targetMethod.Name))
        {
            hasRegisterMethod = true;
        }
        else if (msBuildAssemblies.Contains(targetMethod.ContainingAssembly.Name)))
        {
            hasBadReference = true;
        }
    }, OperationKind.Invocation);

    context.RegisterOperationAction(context =>
    {
        var type = context.Operation switch
        {
            IObjectCreationOperation objectCreation => objectCreation.Type,
            ITypeOfOperation typeofOperation => typeofOperation.TypeOperand,
            _ => null,
        };

        if (msBuildAssemblies.Contains(type?.ContainingAssembly.Name)))
        {
            hasBadReference = true;
        }
    }, OperationKind.ObjectCreation, OperationKind.TypeOf);

    context.RegisterOperationBlockEndAction(context =>
    {
        if (hasRegisterMethod && hasBadReference)
        {
            // report diagnostic
        }
    });
});
```

Not sure if that's better. I wrote directly in GitHub so could have made mistakes, but the idea is to first register operation block start action, then on that, register operation actions for invocation, typeof, and object creation and gather the needed info (whether a register method is called and whether there is another problematic reference)

Then, on operation block end, we have gathered the needed info, and decide whether to report a diagnostic or not.

src/MSBuildLocatorAnalyzer/MSBuildLocatorAnalyzerAnalyzer.cs
```diff
@@ -0,0 +1,75 @@
...
+
+        public override void Initialize(AnalysisContext context)
+        {
+            context.RegisterOperationAction(opact =>
+            {
+                var invocation = (IInvocationOperation)opact.Operation;
+                var targetMethod = invocation.TargetMethod;
+                if (targetMethod?.ContainingAssembly != null)
+                {
+                    if (targetMethod.ContainingAssembly.Name.Equals("Microsoft.Build.Locator") && msbuildLocatorRegisterMethods.Contains(targetMethod.Name))
```
Youssef1313, line 53: Consider having a compilation start action that does:

```csharp
var locatorSymbol = context.Compilation.GetTypeByMetadataName("Microsoft.Build.Locator.MSBuildLocator");
if (locatorSymbol is null)
    return;

// then check targetMethod.ContainingType.Equals(locatorSymbol, SymbolEqualityComparer.Default) instead of comparing the containing assembly name
```

Forgind:
> @Forgind, do you plan to finalize this PR?

I think this was a worthwhile project in getting people to not make this (rather confusing) mistake, but I didn't know how to actually ship the analyzer such that it would run on users' code. I think it might be worthwhile to ship this (and I can take Youssef1313's comments into account if so) if someone can offer a suggestion as to how to make the analyzer run for users?

src/MSBuildLocatorAnalyzer/MSBuildLocatorAnalyzer.csproj
```diff
@@ -0,0 +1,31 @@
...
+	  <RepositoryUrl>https://github.com/microsoft/MSBuildLocator</RepositoryUrl>
+	  <PackageRequiredLicenseAcceptance>false</PackageRequiredLicenseAcceptance>
+	  <Description>Analyzers for MSBuildLocator</Description>
+	  <PackageReleaseNotes>An analyzer that identifies the common problem in MSBuildLocator of registering MSBuild in the same method that uses it.</PackageReleaseNotes>
+	  <Copyright>Copyright</Copyright>
+	  <PackageTags>MSBuildLocator, Build, MSBuild, Analyzers</PackageTags>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.CodeAnalysis.Analyzers" Version="2.9.8" />
```
Youssef1313, line 22: nit: This version is quite old and has known annoying bugs.

src/MSBuildLocatorAnalyzer/MSBuildLocatorAnalyzer.csproj
```diff
@@ -0,0 +1,31 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFrameworks>net48;netcoreapp3.1</TargetFrameworks>
+    <IsPackable>true</IsPackable>
+	<IncludeBuildOutput>false</IncludeBuildOutput>
```
Youssef1313, line 6: This mixes tabs & spaces.

src/MSBuildLocatorAnalyzer/MSBuildLocatorAnalyzer.csproj
```diff
@@ -0,0 +1,31 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFrameworks>net48;netcoreapp3.1</TargetFrameworks>
```
Youssef1313, line 4: This must be netstandard2.0

src/MSBuildLocatorAnalyzer.CodeFixes/MSBuildLocatorAnalyzer.CodeFixes.csproj
```diff
@@ -0,0 +1,37 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFrameworks>net48;netcoreapp3.1</TargetFrameworks>
```
Youssef1313, line 4: This must be netstandard2.0

I missed reviewing this earlier and only looked when you asked for how this is going to be shipped.

Youssef1313 (commented):
Regarding packaging this. You can check the analyzers template:

https://github.com/dotnet/roslyn-sdk/tree/main/src/VisualStudio.Roslyn.SDK/Roslyn.SDK/ProjectTemplates/CSharp/Diagnostic

Basically, we would need a project that packs both the analyzer and codefix dlls in `analyzers/dotnet/cs` directory, let's name the project `MSBuildLocatorAnalyzer.Package` for example. It would be similar to this:

https://github.com/dotnet/roslyn-sdk/blob/main/src/VisualStudio.Roslyn.SDK/Roslyn.SDK/ProjectTemplates/CSharp/Diagnostic/Package/Package.csproj

The analyzers csproj would be as simple as this:

https://github.com/dotnet/roslyn-sdk/blob/main/src/VisualStudio.Roslyn.SDK/Roslyn.SDK/ProjectTemplates/CSharp/Diagnostic/Analyzer/DiagnosticAnalyzer.csproj

as well as the codefix project:

https://github.com/dotnet/roslyn-sdk/blob/main/src/VisualStudio.Roslyn.SDK/Roslyn.SDK/ProjectTemplates/CSharp/Diagnostic/CodeFix/CodeFixProvider.csproj

YuliiaKovalova:
Close with the comment: https://github.com/microsoft/MSBuildLocator/pull/116#issuecomment-1593615775

PR #117, opened 2020-12-29
Fix to return all installed DotNetSDK versions

vxfield:
This PR is a proposed fix/improvements to the following PRs to address [qsharp-compiler#737](https://github.com/microsoft/qsharp-compiler/issues/737):
 - #79 Return all installed DotNetSDK versions
   - Fix console output event handling
     - Problem: if the "dotnet --info" process runs faster and completes before we subscribe to the output events we loose its output. And we also were checking if the process completes successfully before this [line](https://github.com/microsoft/MSBuildLocator/pull/79/files#diff-82450af61239e35c0ab2b9c2fae5c42392a4807ae063b89660e5fdc9fce46c8eR98) we were existing without the actual list of SDKs.
     - Fixes:
       - [Subscribe to output events before starting the process](https://github.com/microsoft/MSBuildLocator/pull/117/files#diff-82450af61239e35c0ab2b9c2fae5c42392a4807ae063b89660e5fdc9fce46c8eR104)
       - [Don't exit if the process has already finished](https://github.com/microsoft/MSBuildLocator/pull/117/files#diff-82450af61239e35c0ab2b9c2fae5c42392a4807ae063b89660e5fdc9fce46c8eL108)
   -  Fix to not require the base path to be present to return the full list of SDKs
      - Problem: the base path is not returned by "dotnet --info" if the required version (from global.json) is not found. Still, we could return the list of installed SDKs and let the caller method to decide whether or not to use the available SDKs.
       - Fix:  [if the base path is found return it, otherwise proceed to list the installed SDKs](https://github.com/microsoft/MSBuildLocator/pull/117/files#diff-82450af61239e35c0ab2b9c2fae5c42392a4807ae063b89660e5fdc9fce46c8eR132)
 - #106 Only consider sdks from before the runtime 
   - Problem: I may be getting this wrong, but my understanding is that it would be ok for a older runtime to use a more recent SDK (assuming backward compatibility) and the opposite (newer SDK using an older runtime) might not be ok. So then, I think [this logic is backward](https://github.com/microsoft/MSBuildLocator/pull/106/files#diff-82450af61239e35c0ab2b9c2fae5c42392a4807ae063b89660e5fdc9fce46c8eR57), because it's exiting when we have a greater version available to be used (major or minor). I'm not sure if this was a mistake or by-design and I'm not getting the problem it solved. But from the Quantum Development Kit perspective, the current logic is returning null when QDK (which is built against 3.1) runs on 5.0 and 3.1 is not available. The correct behavior (for QDK) is to allow 5.0 to be returned as the SDK to be used.
   - Fix: [allow greater versions, block lesser version](https://github.com/microsoft/MSBuildLocator/pull/117/files#diff-82450af61239e35c0ab2b9c2fae5c42392a4807ae063b89660e5fdc9fce46c8eR57)

### DIFF
--- a/src/MSBuildLocator/DotNetSdkLocationHelper.cs
+++ b/src/MSBuildLocator/DotNetSdkLocationHelper.cs
@@ -49,13 +49,13 @@ namespace Microsoft.Build.Locator
             {
                 return null;
             }
-            
+
             // Components of the SDK often have dependencies on the runtime they shipped with, including that several tasks that shipped
             // in the .NET 5 SDK rely on the .NET 5.0 runtime. Assuming the runtime that shipped with a particular SDK has the same version,
             // this ensures that we don't choose an SDK that doesn't work with the runtime of the chosen application. This is not guaranteed
             // to always work but should work for now.
-            if (major > Environment.Version.Major ||
-                (major == Environment.Version.Major && minor > Environment.Version.Minor))
+            if (major < Environment.Version.Major ||
+                (major == Environment.Version.Major && minor < Environment.Version.Minor))
             {
                 return null;
             }
@@ -82,42 +82,40 @@ namespace Microsoft.Build.Locator
             const string DOTNET_CLI_UI_LANGUAGE = nameof(DOTNET_CLI_UI_LANGUAGE);
             
             Process process;
+            var lines = new List<string>();
             try
             {
-                var startInfo = new ProcessStartInfo("dotnet", "--info")
+                process = new Process()
                 {
-                    WorkingDirectory = workingDirectory,
-                    CreateNoWindow = true,
-                    UseShellExecute = false,
-                    RedirectStandardOutput = true,
-                    RedirectStandardError = true
+                    StartInfo = new ProcessStartInfo("dotnet", "--info")
+                    {
+                        WorkingDirectory = workingDirectory,
+                        CreateNoWindow = true,
+                        UseShellExecute = false,
+                        RedirectStandardOutput = true,
+                        RedirectStandardError = true
+                    }
                 };
 
                 // Ensure that we set the DOTNET_CLI_UI_LANGUAGE environment variable to "en-US" before
                 // running 'dotnet --info'. Otherwise, we may get localized results.
-                startInfo.EnvironmentVariables[DOTNET_CLI_UI_LANGUAGE] = "en-US";
+                process.StartInfo.EnvironmentVariables[DOTNET_CLI_UI_LANGUAGE] = "en-US";
 
-                process = Process.Start(startInfo);
+                process.OutputDataReceived += (_, e) =>
+                {
+                    if (!string.IsNullOrWhiteSpace(e.Data))
+                    {
+                        lines.Add(e.Data);
+                    }
+                };
+
+                process.Start();
             }
             catch
             {
                 // when error running dotnet command, consider dotnet as not available
                 yield break;
             }
-
-            if (process.HasExited)
-            {
-                yield break;
-            }
-
-            var lines = new List<string>();
-            process.OutputDataReceived += (_, e) =>
-            {
-                if (!string.IsNullOrWhiteSpace(e.Data))
-                {
-                    lines.Add(e.Data);
-                }
-            };
 
             process.BeginOutputReadLine();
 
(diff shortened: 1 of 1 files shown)


vxfield:
Tagging @rainersigwald, @kurtgronbech, @kurtcodemander, @ladipro, @Forgind 
as you have been working on the fix.

Forgind:
@vxfield,

As I understand it, the SDK has various parts that call into the runtime, so having a newer SDK will only work if the ways the SDK is used happen to not call any code paths that use newly exposed parts of the runtime API. https://github.com/microsoft/qsharp-compiler/issues/737 said the problem emerged "after installing the .NET 5 SDK," which prompted MSBuildLocator to choose the .NET 5 SDK over 3.1. The proposed workaround—specifying 3.1 in global.json—again just meant the SDK version didn't exceed the runtime version.

I think the problem is in adding an extra step. My PR was meant to solve a problem in which project A builds against 3.1 and uses MSBuildLocator to look for an appropriate MSBuild to do some part of the work. In that case, the work may fail if it finds the 5.0 SDK because, as I mentioned, the 5.0 SDK may use parts of the 5.0 runtime that aren't loaded. If project B uses A.dll and components from the 5.0 SDK that aren't in the 3.1 SDK, can you build that against 5.0? And does B need MSBuildLocator?

Is is possible to build everything against 5.0? It might also be reasonable to use RegisterMSBuildPath or RegisterInstance to get more precision with which version you want.

src/MSBuildLocator/DotNetSdkLocationHelper.cs
```diff
@@ -82,58 +82,58 @@ private static IEnumerable<string> GetDotNetBasePaths(string workingDirectory)
             const string DOTNET_CLI_UI_LANGUAGE = nameof(DOTNET_CLI_UI_LANGUAGE);
             
             Process process;
+            var lines = new List<string>();
             try
             {
-                var startInfo = new ProcessStartInfo("dotnet", "--info")
+                process = new Process()
```
rainersigwald, line 88: Is this change necessary?

src/MSBuildLocator/DotNetSdkLocationHelper.cs
```diff
@@ -82,58 +82,58 @@ private static IEnumerable<string> GetDotNetBasePaths(string workingDirectory)
...
-                process = Process.Start(startInfo);
+                process.OutputDataReceived += (_, e) =>
+                {
+                    if (!string.IsNullOrWhiteSpace(e.Data))
+                    {
+                        lines.Add(e.Data);
+                    }
+                };
+
+                process.Start();
```
rainersigwald, line 112: This reordering and moving: 👍🏻

src/MSBuildLocator/DotNetSdkLocationHelper.cs
```diff
@@ -82,58 +82,58 @@ private static IEnumerable<string> GetDotNetBasePaths(string workingDirectory)
...
+            // FirstOrDefault always returns the version in use.
+            // The BasePath might not be found if the required version (via global.json)
+            // is not installed. In that case, we can still iterate to the other SDKs installed
+            string basePath = null;
+            if (matched.Success)
             {
-                yield break; 
+                basePath = matched.Groups[1].Value.Trim();
+                yield return basePath; 
             }
```
rainersigwald, line 136: This looks good too.

src/MSBuildLocator/DotNetSdkLocationHelper.cs
```diff
@@ -49,13 +49,13 @@ public static VisualStudioInstance GetInstance(string dotNetSdkPath)
...
-            
+
             // Components of the SDK often have dependencies on the runtime they shipped with, including that several tasks that shipped
             // in the .NET 5 SDK rely on the .NET 5.0 runtime. Assuming the runtime that shipped with a particular SDK has the same version,
             // this ensures that we don't choose an SDK that doesn't work with the runtime of the chosen application. This is not guaranteed
             // to always work but should work for now.
-            if (major > Environment.Version.Major ||
-                (major == Environment.Version.Major && minor > Environment.Version.Minor))
+            if (major < Environment.Version.Major ||
+                (major == Environment.Version.Major && minor < Environment.Version.Minor))
```
rainersigwald, line 58: This was correct before.

The .NET SDK is build concurrently with the .NET runtime, and SDK features may use new runtime features. The .NET 5.0 SDK _must_ execute on the .NET 5.0 runtime. We expect that it will also run fine on .NET 6.0 and higher, barring surprising .NET runtime breaking changes.

rainersigwald (changes requested):
> But from the Quantum Development Kit perspective, the current logic is returning null when QDK (which is built against 3.1) runs on 5.0 and 3.1 is not available. The correct behavior (for QDK) is to allow 5.0 to be returned as the SDK to be used.

That was the previous behavior of MSBuildLocator, which was the cause of microsoft/qsharp-compiler#737: you tried to load parts of the 5.0 SDK in-proc and failed, because they had dependencies on the 5.0 runtime.

If you continue to target .NET Core 3.1, you will not be able to use the .NET 5.0 SDK--the only path forward there is to target 5.0 yourselves. That should work fine even if a user uses the 3.1 SDK (because the .NET runtime is generally highly backward compatible).

rainersigwald:
As a style nitpick, the PR description is a great walkthrough of what changed--thanks! But we'd prefer it if you put those individual changes in individual commits. That would also make it easier to drop the one change here that's incorrect.

vxfield:
Thank you, @forgind and @rainersigwald, for the comments and clarifications.
We are evaluating moving the Q# Compiler to build against 5.0 per your recommendation.

I'm cancelling/closing this PR and created another #119 to just address some reliability improvements to the MS Build Locator listing all available SDKs.

PR #118, opened 2021-01-13
Only look in extra NuGet directory from VS install

Forgind:
I neglected to test the RegisterMSBuildPath to a not-VS location scenario, and it can fail. This should fix it.

### DIFF
--- a/src/MSBuildLocator/MSBuildLocator.cs
+++ b/src/MSBuildLocator/MSBuildLocator.cs
@@ -124,7 +124,7 @@ namespace Microsoft.Build.Locator
             }
 
             // Find and load NuGet assemblies if msbuildPath is in a VS installation
-            string nugetPath = Path.GetFullPath(Path.Combine(instance.MSBuildPath, "..", "..", "..", "Common7", "IDE", "CommonExtensions", "Microsoft", "NuGet"))
+            string nugetPath = Path.GetFullPath(Path.Combine(instance.MSBuildPath, "..", "..", "..", "Common7", "IDE", "CommonExtensions", "Microsoft", "NuGet"));
             if (Directory.Exists(nugetPath))
             {
                 RegisterMSBuildPath(new string[] { instance.MSBuildPath, nugetPath });
--- a/src/MSBuildLocator/MSBuildLocator.cs
+++ b/src/MSBuildLocator/MSBuildLocator.cs
@@ -123,7 +123,19 @@ namespace Microsoft.Build.Locator
                 ApplyDotNetSdkEnvironmentVariables(instance.MSBuildPath);
             }
 
-            RegisterMSBuildPath(instance.MSBuildPath);
+            if (instance.DiscoveryType == DiscoveryType.VisualStudioSetup || instance.DiscoveryType == DiscoveryType.DeveloperConsole)
+            {
+                RegisterMSBuildPath(new string[]
+                {
+                    instance.MSBuildPath,
+                    // Find and load NuGet assemblies if msbuildPath is in a VS installation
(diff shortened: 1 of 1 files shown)


src/MSBuildLocator/MSBuildLocator.cs
```diff
@@ -123,7 +123,19 @@ public static void RegisterInstance(VisualStudioInstance instance)
...
             }
 
-            RegisterMSBuildPath(instance.MSBuildPath);
+            if (instance.DiscoveryType == DiscoveryType.VisualStudioSetup || instance.DiscoveryType == DiscoveryType.DeveloperConsole)
+            {
+                RegisterMSBuildPath(new string[]
+                {
+                    instance.MSBuildPath,
+                    // Find and load NuGet assemblies if msbuildPath is in a VS installation
+                    Path.GetFullPath(Path.Combine(instance.MSBuildPath, "..", "..", "..", "Common7", "IDE", "CommonExtensions", "Microsoft", "NuGet"))
```
rainersigwald, line 132: Should this be conditional on it existing?

rainersigwald: Specifically we probably want this for corext type discovery too.

Forgind: That was the problem with the previous spot, but it should always exist if we're in VS or a developer console, right?

Forgind: Corext type discovery? I only see three options (pre-mono): developer console, VS, and dotnet sdk.

rainersigwald: #84. I think it'd be easiest to do the exists check but we could extend the if there as well.

src/MSBuildLocator/MSBuildLocator.cs
```diff
@@ -123,7 +123,19 @@ public static void RegisterInstance(VisualStudioInstance instance)
...
+                });
+            }
```
Forgind, line 134:
```suggestion
            // Find and load NuGet assemblies if msbuildPath is in a VS installation
            string nugetPath = Path.GetFullPath(Path.Combine(instance.MSBuildPath, "..", "..", "..", "Common7", "IDE", "CommonExtensions", "Microsoft", "NuGet"))
            if (Directory.Exists(nugetPath))
            {
                RegisterMSBuildPath(new string[] { instance.MSBuildPath, nugetPath });
            }
```

src/MSBuildLocator/MSBuildLocator.cs
Forgind: Good point—forgot about that PR. Fixed!

src/MSBuildLocator/MSBuildLocator.cs
```diff
@@ -123,7 +123,16 @@ public static void RegisterInstance(VisualStudioInstance instance)
                 ApplyDotNetSdkEnvironmentVariables(instance.MSBuildPath);
             }
 
-            RegisterMSBuildPath(instance.MSBuildPath);
+            // Find and load NuGet assemblies if msbuildPath is in a VS installation
+            string nugetPath = Path.GetFullPath(Path.Combine(instance.MSBuildPath, "..", "..", "..", "Common7", "IDE", "CommonExtensions", "Microsoft", "NuGet"))
```
Forgind, line 127:
```suggestion
            string nugetPath = Path.GetFullPath(Path.Combine(instance.MSBuildPath, "..", "..", "..", "Common7", "IDE", "CommonExtensions", "Microsoft", "NuGet"));
```

rainersigwald (approved):

PR #121, opened 2021-01-20
Update version to 1.4

Forgind:
Is there any reason we aren't updating the assembly version?

### DIFF
--- a/version.json
+++ b/version.json
@@ -1,5 +1,5 @@
 {
-  "version": "1.3",
+  "version": "1.4",
   "assemblyVersion": "1.0.0.0",
   "publicReleaseRefSpec": [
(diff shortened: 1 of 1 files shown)


cdmihai (approved):

jeffkl:
> Is there any reason we aren't updating the assembly version?

There isn't really a reason to update it unless you'd want users to be able to load two distinct versions.  Updating the assembly versions causes all kinds of headaches around binding redirects and dependency chains.  Leaving it as 1.0.0.0 forever means that any assembly compiled against any version will get whatever version the main application wants and doesn't have to deal with all of that.

jeffkl (approved):

PR #122, opened 2021-01-29
Add better build defaults

AndyGerlicher:


Directory.Build.rsp
```diff
@@ -0,0 +1,6 @@
+-Restore
+-ConsoleLoggerParameters:Verbosity=Minimal;Summary;ForceNoAlign
+-MaxCPUCount
+-NodeReuse:false
+-graph
+-p:RestoreUseStaticGraphEvaluation=true
```
jeffkl, line 6: Let's use Directory.Solution.props instead

AndyGerlicher: Only downside here is you have to but it in both `Directory.Solution.props` (for when you build the sln) and `Directory.Build.props` (for when you build something else) 😞

jeffkl: It’s actually good because d.b.props is imported for every project in the tree and d.s.props is imported for only solution files. If it wasn’t this way, you’d have to use conditions everywhere for logic you only want to apply to projects vs a solution. Solutions build very differently. Not to mention having solutions import d.b.props all of a sudden would have been a breaking change. So it is what it is 😀

jeffkl (approved):

Directory.Build.rsp
jzabroski: MSBuild is a giant fancy if statement that nobody can write down, :)

PR #129, opened 2021-05-13
Don't read stderr when running .NET CLI

### LINKED ISSUE #128: MSBuildLocator.QueryVisualStudioInstances hangs if COREHOST_TRACE is set to 1
A hang can occur when locating the .NET SDK using `dotnet --info`, if the `COREHOST_TRACE` environment variable is set to `1`. This is because [RedirectStandardError is set to `true`](https://github.com/microsoft/MSBuildLocator/blob/master/src/MSBuildLocator/DotNetSdkLocationHelper.cs#L96), and nothing reads from the standard error stream.

FiniteReality:
Fixes #128


rainersigwald (approved):
Thanks!

PR #132, opened 2021-05-19
Enable all searches on any framework, with sane defaults

### LINKED ISSUE #131: Allow using VisualStudioLocationHelper on .NET Core
I'm working on https://github.com/dotnet/upgrade-assistant that runs as a .NET 5 app, but we need to be able to get msbuild location from VS to find some targets such as WebApplications.targets. However, we end up having to duplicate the VisualStudioLocationHelper so that we can use it. Can this be exposed as an option for query? I totally understand not having it be included by default, but would love to be able to opt into it.

twsouthwick:
This change enables apps running under .NET Core to find instances of VS/DevCmd prompt if needed. On .NET Framework, it allows searching for .NET Core. It does not change the defaults, though. 

Fixes #131

### DIFF
--- a/src/MSBuildLocator/DotNetSdkLocationHelper.cs
+++ b/src/MSBuildLocator/DotNetSdkLocationHelper.cs
@@ -1,8 +1,6 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
-#if NETCOREAPP
-
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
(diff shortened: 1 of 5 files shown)


twsouthwick:
@rainersigwald Is this something we could get merged in?

teo-tsirpanis:
Won't it cause problems if a .NET Framework app loads a .NET Core assembly and vice versa?

jzabroski:
@rainersigwald We need to stop the madness with MSBuildLocator and people creating crazy look-up rules with no tests.

teo-tsirpanis:
> Yes. I was using it more for the locator aspect [...]

Then I think we should also check and throw if we try to load a `VisualStudioInstance` with a `DiscoveryType` equal or unequal to `DotNetSdk`, when we are on .NET Framework and .NET Core respectively.

twsouthwick:
This doesn't help people who need to load the assemblies. This is helpful, though, for the use case of finding targets the VS installs. However, I'm happy to just use the VS interop directly as it sounds like that is not the main use case of this tool

PR #135, opened 2021-09-20
Find x64 MSBuild from Locator

Forgind:
Fixes https://github.com/dotnet/msbuild/issues/6681 for pre-17.0 MSBuild.

### DIFF
--- a/src/MSBuildLocator/MSBuildLocator.cs
+++ b/src/MSBuildLocator/MSBuildLocator.cs
@@ -209,6 +209,27 @@ namespace Microsoft.Build.Locator
             // AssemblyResolve event can fire multiple times for the same assembly, so keep track of what's already been loaded.
             var loadedAssemblies = new Dictionary<string, Assembly>();
 
+            // MSBuild can be loaded from the x86 or x64 folder. Before 17.0, it looked next to the executing assembly in some cases and constructed a path that assumed x86 in others.
+            // This overrides the latter assumption to let it find the right MSBuild.
+            foreach (string path in msbuildSearchPaths)
+            {
+                string msbuildExe = Path.Combine(path, "MSBuild.exe");
+                if (File.Exists(msbuildExe))
+                {
+                    Environment.SetEnvironmentVariable("MSBUILD_EXE_PATH", msbuildExe);
+                    break;
+                }
+                else
+                {
+                    string msbuildDll = Path.Combine(path, "MSBuild.dll");
+                    if (File.Exists(msbuildDll))
+                    {
+                        Environment.SetEnvironmentVariable("MSBUILD_EXE_PATH", msbuildDll);
+                        break;
+                    }
+                }
+            }
+
             // Saving the handler in a static field so it can be unregistered later.
 #if NET46
             s_registeredHandler = (_, eventArgs) =>


src/MSBuildLocator/MSBuildLocator.cs
```diff
@@ -209,6 +209,27 @@ public static void RegisterMSBuildPath(string[] msbuildSearchPaths)
             // AssemblyResolve event can fire multiple times for the same assembly, so keep track of what's already been loaded.
             var loadedAssemblies = new Dictionary<string, Assembly>();
 
+            // MSBuild can be loaded from the x86 or x64 folder. Before 17.0, it looked next to the executing assembly in some cases and constructed a path that assumed x86 in others.
+            // This overrides the latter assumption to let it find the right MSBuild.
+            foreach (string path in msbuildSearchPaths)
```
rainersigwald, line 214: Do we not need to do the `amd64` folder removal dance here too, in case someone passed in a 64-bit folder?

Can we do this only for <17.0?

rainersigwald, line 214: Do we not need to do the `amd64` folder removal dance here too, in case someone passed in a 64-bit folder?

Can we do this only for <17.0?

src/MSBuildLocator/MSBuildLocator.cs
```diff
@@ -209,6 +209,27 @@ public static void RegisterMSBuildPath(string[] msbuildSearchPaths)
...
+            {
+                string msbuildExe = Path.Combine(path, "MSBuild.exe");
+                if (File.Exists(msbuildExe))
+                {
+                    Environment.SetEnvironmentVariable("MSBUILD_EXE_PATH", msbuildExe);
+                    break;
+                }
+                else
+                {
+                    string msbuildDll = Path.Combine(path, "MSBuild.dll");
```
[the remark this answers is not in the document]
Forgind: I wasn't 100% sure of this, but if you have a .NET Framework app and use RegisterMSBuildPath(<.NET Core MSBuild path>), I think you would have trouble if you tried to use anything that differs between Core and Framework, but you could maybe stick to the overlap and having a working app. Should that be something to care about?

src/MSBuildLocator/MSBuildLocator.cs
Forgind: Isn't that handled in MSBuild? I'm looking at TryFromMSBuildExeUnderVisualStudio, and it calculates visualStudioRoot by checking whether there's an amd64 and going one extra level up if so.

On the other path, it assumes a standalone MSBuild.exe, so I wasn't sure what assumptions we could actually make about it.

Forgind: Doing it for < 17.0 sounds reasonable. Little more expensive perf-wise, but that isn't a big deal in Locator.

rainersigwald: Are you looking at MSBuild 16 or 17 source? That was a thing we fixed, right?

src/MSBuildLocator/MSBuildLocator.cs
```diff
@@ -209,6 +209,25 @@ public static void RegisterMSBuildPath(string[] msbuildSearchPaths)
...
             var loadedAssemblies = new Dictionary<string, Assembly>();
 
+            // MSBuild can be loaded from the x86 or x64 folder. Before 17.0, it looked next to the executing assembly in some cases and constructed a path that assumed x86 in others.
+            // This overrides the latter assumption to let it find the right MSBuild.
+            foreach (string path in msbuildSearchPaths)
+            {
+                string msbuildExe = Path.Combine(path, "MSBuild.exe");
+                if (File.Exists(msbuildExe))
+                {
+                    if (FileVersionInfo.GetVersionInfo(msbuildExe).FileMajorPart < 17)
```
rainersigwald, line 219: Should we do an exact check on the subversion we introduced the change in?

Forgind: It came in in 17.0, so the minor version wouldn't matter. We could make sure it isn't a preview before it was merged, but feels like more work than is warranted. Not a ton of people use previews, and if they're happy being on one preview, they can update to a later one.

PR #147, opened 2021-11-04
Prolong version check

Forgind:
The change in #135 was supposed to be made unnecessary by https://github.com/dotnet/msbuild/pull/6890, but apparently my setup is the only one that has dotnet.exe under both the dotnet folder and the sdks folder, so it didn't work. xen2 fixed the problem in https://github.com/dotnet/msbuild/pull/7013, but that won't get in 'til 17.1, so we need to prolong the version check here.

### DIFF
--- a/src/MSBuildLocator/MSBuildLocator.cs
+++ b/src/MSBuildLocator/MSBuildLocator.cs
@@ -218,7 +218,7 @@ namespace Microsoft.Build.Locator
                 if (File.Exists(msbuildExe))
                 {
                     FileVersionInfo ver = FileVersionInfo.GetVersionInfo(msbuildExe);
-                    if (ver.FileMajorPart < 17 || ver.FileMinorPart < 1)
+                    if (ver.FileMajorPart < 17 || (ver.FileMajorPart == 17 && ver.FileMinorPart < 1))
                     {
                         if (Path.GetDirectoryName(msbuildExe).EndsWith(@"\amd64", StringComparison.OrdinalIgnoreCase))
                         {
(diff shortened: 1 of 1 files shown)


src/MSBuildLocator/MSBuildLocator.cs
```diff
@@ -217,7 +217,8 @@ public static void RegisterMSBuildPath(string[] msbuildSearchPaths)
                 string msbuildExe = Path.Combine(path, "MSBuild.exe");
                 if (File.Exists(msbuildExe))
                 {
-                    if (FileVersionInfo.GetVersionInfo(msbuildExe).FileMajorPart < 17)
+                    FileVersionInfo ver = FileVersionInfo.GetVersionInfo(msbuildExe);
+                    if (ver.FileMajorPart < 17 || ver.FileMinorPart < 1)
```
xen2, line 221: Is that OK? Won't this fail for future 18.0?

xen2: In that case, shouldn't the check be `if (ver.FileMajorPart < 17 || (ver.FileMajorPart == 17 && ver.FileMinorPart < 1))`?
Otherwise it will still set the variables for 18.0 since `(ver.FileMinorPart < 1)` is true.

Forgind: 🤦‍♂️

Yes. Thank you; you've been a big help with this.

Forgind:
```suggestion
                    if (ver.FileMajorPart < 17 || (ver.FileMajorPart == 17 && ver.FileMinorPart < 1))
```

Jonhi1393:
> The change in #135 was supposed to be made unnecessary by https://github.com/dotnet/msbuild/pull/6890, but apparently my setup is the only one that has dotnet.exe under both the dotnet folder and the sdks folder, so it didn't work. xen2 fixed the problem in https://github.com/dotnet/msbuild/pull/7013, but that won't get in 'til 17.1, so we need to prolong the version check here.

Forgind:
Hi @Jonhi1393,

You quoted my message but didn't add anything. What were you trying to say?

PR #161, opened 2022-06-23
Use hostfxr hosting layer API

Forgind:
This makes the process of finding .NET SDK instances more efficient and fix dotnet/msbuild#7466

### DIFF
--- a/src/MSBuildLocator/DotNetSdkLocationHelper.cs
+++ b/src/MSBuildLocator/DotNetSdkLocationHelper.cs
@@ -78,91 +78,44 @@ namespace Microsoft.Build.Locator
             }
         }
 
-        enum hostfxr_resolve_sdk2_flags_t
-        {
-            disallow_prerelease = 0x1,
-        };
-
-        enum hostfxr_resolve_sdk2_result_key_t
-        {
-            resolved_sdk_dir = 0,
-            global_json_path = 1,
-        };
-
-        [UnmanagedFunctionPointer(CallingConvention.Cdecl, CharSet = CharSet.Auto)]
-        private delegate void hostfxr_resolve_sdk2_result_fn(
-                hostfxr_resolve_sdk2_result_key_t key,
-                string value);
-
-        [UnmanagedFunctionPointer(CallingConvention.Cdecl, CharSet = CharSet.Auto)]
-        private delegate void hostfxr_get_available_sdks_result_fn(
-                hostfxr_resolve_sdk2_result_key_t key,
-                [MarshalAs(UnmanagedType.LPArray, SizeParamIndex = 0)]
-                string[] value);
-
-        [DllImport("hostfxr", CharSet = CharSet.Auto, ExactSpelling = true, CallingConvention = CallingConvention.Cdecl)]
-        private static extern int hostfxr_resolve_sdk2(
-            string exe_dir,
-            string working_dir,
-            hostfxr_resolve_sdk2_flags_t flags,
-            hostfxr_resolve_sdk2_result_fn result);
-
-        [DllImport("hostfxr", CharSet = CharSet.Auto, ExactSpelling = true, CallingConvention = CallingConvention.Cdecl)]
-        private static extern int hostfxr_get_available_sdks(string exe_dir, hostfxr_get_available_sdks_result_fn result);
-
-        [DllImport("libc", ExactSpelling = true, CallingConvention = CallingConvention.Cdecl)]
-        private static extern IntPtr realpath(string path, IntPtr buffer);
-
-        [DllImport("libc", ExactSpelling = true, CallingConvention = CallingConvention.Cdecl)]
-        private static extern void free(IntPtr ptr);
-
         private static string realpath(string path)
         {
-            IntPtr ptr = realpath(path, IntPtr.Zero);
+            IntPtr ptr = NativeMethods.realpath(path, IntPtr.Zero);
             string result = Marshal.PtrToStringAuto(ptr);
-            free(ptr);
+            NativeMethods.free(ptr);
             return result;
         }
 
         private static IEnumerable<string> GetDotNetBasePaths(string workingDirectory)
         {
             string dotnetPath = null;
+            bool isWindows = RuntimeInformation.IsOSPlatform(OSPlatform.Windows);
+            string exeName = isWindows ? "dotnet.exe" : "dotnet";
 
-            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+            // We will generally find the dotnet exe on the path, but on linux, it is often just a 'dotnet' symlink (possibly even to more symlinks) that we have to resolve
+            // to the real dotnet executable.
+            // This will work as often as just invoking dotnet from the command line, but we can be more confident in finding a dotnet executable by following
+            // https://github.com/dotnet/designs/blob/main/accepted/2021/install-location-per-architecture.md
+            // This can be done using the nethost library. We didn't do this previously, so I did not implement this extension.
+            foreach (string dir in Environment.GetEnvironmentVariable("PATH").Split(Path.PathSeparator))
             {
-                foreach (string dir in Environment.GetEnvironmentVariable("PATH").Split(';'))
+                string filePath = Path.Combine(dir, exeName);
+                if (File.Exists(Path.Combine(dir, exeName)))
                 {
-                    if (File.Exists(Path.Combine(dir, "dotnet.exe")))
-                    {
-                        dotnetPath = dir;
-                        break;
-                    }
-                }
-            }
-            else
-            {
-                foreach (string dir in Environment.GetEnvironmentVariable("PATH").Split(':'))
-                {
-                    string filePath = Path.Combine(dir, "dotnet");
-                    if (File.Exists(filePath))
-                    {
-                        dotnetPath = filePath;
-                        break;
-                    }
-                }
-
-                if (dotnetPath != null)
-                {
-                    dotnetPath = realpath(dotnetPath) ?? dotnetPath;
-                    dotnetPath = Path.GetDirectoryName(dotnetPath);
+                    dotnetPath = filePath;
+                    break;
                 }
             }
 
+            if (dotnetPath != null)
+            {
+                dotnetPath = Path.GetDirectoryName(isWindows ? dotnetPath : realpath(dotnetPath) ?? dotnetPath);
+            }
 
             string bestSDK = null;
-            int rc = hostfxr_resolve_sdk2(exe_dir: dotnetPath, working_dir: workingDirectory, flags: 0, result: (key, value) =>
+            int rc = NativeMethods.hostfxr_resolve_sdk2(exe_dir: dotnetPath, working_dir: workingDirectory, flags: 0, result: (key, value) =>
             {
-                if (key == hostfxr_resolve_sdk2_result_key_t.resolved_sdk_dir)
+                if (key == NativeMethods.hostfxr_resolve_sdk2_result_key_t.resolved_sdk_dir)
                 {
                     bestSDK = value;
                 }
@@ -174,7 +127,7 @@ namespace Microsoft.Build.Locator
             }
 
             string[] paths = null;
-            hostfxr_get_available_sdks(exe_dir: dotnetPath, result: (key, value) =>
+            NativeMethods.hostfxr_get_available_sdks(exe_dir: dotnetPath, result: (key, value) =>
             {
                 paths = value;
             });
--- a/src/MSBuildLocator/DotNetSdkLocationHelper.cs
+++ b/src/MSBuildLocator/DotNetSdkLocationHelper.cs
@@ -142,7 +142,10 @@ namespace Microsoft.Build.Locator
             string bestSDK = null;
             int rc = hostfxr_resolve_sdk2(exe_dir: dotnetPath, working_dir: workingDirectory, flags: 0, result: (key, value) =>
             {
-                bestSDK = value;
+                if (key == hostfxr_resolve_sdk2_result_key_t.resolved_sdk_dir)
+                {
+                    bestSDK = value;
+                }
             });
 
             if (rc == 0 && bestSDK != null)
--- a/src/MSBuildLocator/DotNetSdkLocationHelper.cs
+++ b/src/MSBuildLocator/DotNetSdkLocationHelper.cs
@@ -128,19 +128,19 @@ namespace Microsoft.Build.Locator
         {
             string dotnetPath = null;
 
-            // Windows
-            foreach (string dir in Environment.GetEnvironmentVariable("PATH").Split(';'))
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
             {
-                if (File.Exists(Path.Combine(dir, "dotnet.exe")))
+                foreach (string dir in Environment.GetEnvironmentVariable("PATH").Split(';'))
                 {
-                    dotnetPath = dir;
-                    break;
+                    if (File.Exists(Path.Combine(dir, "dotnet.exe")))
+                    {
+                        dotnetPath = dir;
+                        break;
+                    }
                 }
             }
-
-            if (dotnetPath is null)
+            else
             {
-                // Unix
                 foreach (string dir in Environment.GetEnvironmentVariable("PATH").Split(':'))
                 {
                     string filePath = Path.Combine(dir, "dotnet");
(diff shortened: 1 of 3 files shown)


jzabroski:
@rainersigwald I remain perplexed how you can review and/or test quality assurance of any of these changes.

Forgind:
https://github.com/dotnet/sdk/blob/e3c0cf07006be3eb3d66e84125bb03a967672c33/src/Resolvers/Microsoft.DotNet.NativeWrapper/Interop.cs#L132-L201

jzabroski:
Hi @forgind

I'm not trying to negatively criticize your improvements here. I just see a lot of changes to MSBuildLocator with no tests. There is a huge pile of assumptions that underlies how MSBuildLocator even works that isn't routinely proven and/or is legacy.

For example, if the goal of this commit is to improve performance, it would be nice to benchmark that.

I'm curious what other build tools like Bazel do for stuff like this.

Forgind:
@jzabroski,

You're absolutely right that the lack of strong testing infrastructure is a major issue in this repo. My question is: how would you test it? Take this change, for example. This changes how we first find a list of sdks that might have MSBuild. We could try to mock the native call and verify that the mocked native call returns as expected, but that's really just testing the test, not testing the change. We could download a specific set of SDKs and verify that those and only those are found, but that's an end-to-end test, not a unit test; having a long list of those would be very expensive computationally. We could have a short list of end-to-end tests, but at that point, how different is it from manually testing a change?

If you have a good idea for how to write unit tests for MSBuildLocator or feel moved to write a few solid end-to-end tests, we would likely take them. Short of that, I think manual testing is sufficient. This is only issue/pr#161; if it would take our team a lot of effort to create good tests, I don't think it's worth it unless MSBuildLocator suddenly sees a lot more activity.

jzabroski:
> how would you test it?

Docker.

> Short of that, I think manual testing is sufficient.

I don't even know _how_ you manual testing this stuff.  Where is the testing script?

jzabroski:
To elaborate a bit more, there are two open source projects that allow creating ad-hoc docker containers using the Docker API:

1. https://github.com/Deffiss/testenvironment-docker -> https://www.nuget.org/packages/TestEnvironment.Docker
2. https://github.com/testcontainers/testcontainers-dotnet -> https://www.nuget.org/packages/DotNet.Testcontainers

I prefer the first one. I tried to get involved in the second one, but I found it was fairly difficult to communicate with the maintainers. TestContainers is a broader initiative supporting many different languages, so I imagine some of the change control process involves fitting the overall project's architecture. Each supported language doesn't really simply do a language binding, either, but a ground-up native re-implementation.

Despite the fact Deffiss's TestEnvironment.Docker project has about 4x fewer downloads, I think it is a lot cleaner, especially how they take advantage of C# 9 records' `with` syntax. The main disadvantage relative to TestContainers.Net is that Deffiss's project doesn't currently support Moby Ryuk for "resource reaping" (garbage collection for Docker test artifacts like images and containers) .  In general, that shouldn't be that big of a deal.

Windows also now supports podman, so you can use that in some scenarios, but Docker is still better supported, I believe.

Cheers.

Forgind:
Docker is certainly better than what I'd originally been imagining. It would let you run tests with different environments in a fairly clean way. On the other hand, there are still only two relevant parts of MSBuildLocator: does it find the MSBuilds you expect, and after you've picked one, does it let you use the API as expected. Using this change as an example, only the first half is relevant, so if I had great tests for it, they'd create an environment with some SDKs and verify that those SDKs are found. I'd also want to make sure that works on both linux and windows. So that's what I tested. I have several SDKs on my windows computer, and I made sure it found the ones I expected it to find and didn't find the ones I didn't expect it to find. I verified that it put the right one at the top, i.e., sorted by version in descending order. Then I tested it on Ubuntu. That testing is why I noticed the PATH variable is different on windows and Ubuntu, hence the latest change. I also noticed that the call to hostfxr_resolve_sdk_2 returned the path to the global.json rather than the path to the best MSBuild, though I haven't figured out an ideal solution to that yet.

The problem is: there aren't very many tests I can run for MSBuildLocator, and I do run those every time I make a change because they build off of each other. I just run them manually rather than automatically. As I said, I'd love to have it all automated; I'm just not sure if it's worth the time at the moment, given everything else we have to do.

Forgind:
Edit: figured out that last part 🙂

jzabroski:
I think it's pretty easy to create an image that installs a set of SDKs, and associate that image with negative/positive test cases.  Certainly about as much time as manually checking it.  I also think it would be cleaner in terms of catching issues, like returning multiple matches or duplicate matches.

Why do you think it would take longer?

I do agree, the initial investment/learning curve is a bit annoying, but once the guidewire architecture for writing tests is in place, the tests write themselves practically.

src/MSBuildLocator/DotNetSdkLocationHelper.cs
```diff
@@ -77,92 +78,102 @@ public static IEnumerable<VisualStudioInstance> GetInstances(string workingDirec
...
-            const string DOTNET_CLI_UI_LANGUAGE = nameof(DOTNET_CLI_UI_LANGUAGE);
-            
-            Process process;
-            var lines = new List<string>();
-            try
+            string dotnetPath = File.Exists("/usr/share/dotnet/dotnet") ? "/usr/share/dotnet" :
+            File.Exists($"/home/{Environment.GetEnvironmentVariable("USER")}/share/dotnet/dotnet") ? $"/home/{Environment.GetEnvironmentVariable("USER")}/share/dotnet" :
+            File.Exists("dotnet") ? "." :
+            !string.IsNullOrEmpty(Environment.GetEnvironmentVariable("DOTNET_HOME")) ? Environment.GetEnvironmentVariable("DOTNET_HOME") :
+            null;
```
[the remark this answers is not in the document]
Forgind: I asked about the recommended way to find dotnet. I believe your test got different results because the old version skipped to just checking your PATH, whereas this new version looks at several other places that seem more reasonable (if dotnet is there) than PATH. Those other places don't know about this random SDK you just installed but do know about the SDKs you installed the normal way. This is not a case in which I think we should be optimizing for bad customers.

src/MSBuildLocator/DotNetSdkLocationHelper.cs
```diff
@@ -77,92 +78,102 @@ public static IEnumerable<VisualStudioInstance> GetInstances(string workingDirec
...
                     }
-                };
-
-                process.Start();
-            }
-            catch
-            {
-                // when error running dotnet command, consider dotnet as not available
-                yield break;
+                }
```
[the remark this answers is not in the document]
Forgind: I asked. I don't think unifying those makes it better. There are three differences between unix and windows here: Path.PathSeparator, the full executable name, and the delimiter in the PATH variable. Having three places each doing those double checks seems much messier to me, given how little code this actually is. Note that we can't split by both PATH delimiters, as that would break up, for instance, C:\anything

src/MSBuildLocator/DotNetSdkLocationHelper.cs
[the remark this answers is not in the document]
rainersigwald: I think it makes sense to replicate "the build that you would get if you typed `dotnet build` instead of `toolThatUsesLocator`". That would respect `PATH` and multi-level lookup and friends and would not bypass them to find a machine-installed SDK.

src/MSBuildLocator/DotNetSdkLocationHelper.cs
```diff
@@ -77,92 +78,102 @@ public static IEnumerable<VisualStudioInstance> GetInstances(string workingDirec
...
 
-                process.OutputDataReceived += (_, e) =>
+                // Unix
+                foreach (string dir in Environment.GetEnvironmentVariable("PATH").Split(':'))
                 {
-                    if (!string.IsNullOrWhiteSpace(e.Data))
+                    if (File.Exists(Path.Combine(dir, "dotnet")))
                     {
-                        lines.Add(e.Data);
+                        dotnetPath = dir;
```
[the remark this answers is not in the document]
Forgind: No. realpath is for fixing character widths, and CharSet.Auto seems to have taken care of that. I did test this on both windows and WSL, and it worked for both.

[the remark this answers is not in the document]
rainersigwald: > realpath is for fixing character widths

What gave you that impression? `realpath` is for resolving symlinks to their final destination: https://www.man7.org/linux/man-pages/man3/realpath.3.html

src/MSBuildLocator/DotNetSdkLocationHelper.cs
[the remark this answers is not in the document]
jzabroski: > That would respect PATH and multi-level lookup and friends and would not bypass them to find a machine-installed SDK.

That would be a good integration test 👍

src/MSBuildLocator/DotNetSdkLocationHelper.cs
```diff
@@ -77,92 +78,122 @@ public static IEnumerable<VisualStudioInstance> GetInstances(string workingDirec
...
-                        RedirectStandardOutput = true
-                    }
-                };
+            string dotnetPath = null;
 
-                // Ensure that we set the DOTNET_CLI_UI_LANGUAGE environment variable to "en-US" before
-                // running 'dotnet --info'. Otherwise, we may get localized results.
-                process.StartInfo.EnvironmentVariables[DOTNET_CLI_UI_LANGUAGE] = "en-US";
+            // Windows
+            foreach (string dir in Environment.GetEnvironmentVariable("PATH").Split(';'))
```
Forgind, line 132: RuntimeInformation.IsOSPlatform

src/MSBuildLocator/DotNetSdkLocationHelper.cs
```diff
@@ -77,92 +78,122 @@ public static IEnumerable<VisualStudioInstance> GetInstances(string workingDirec
...
+        [DllImport("libc", ExactSpelling = true, CallingConvention = CallingConvention.Cdecl)]
+        private static extern void free(IntPtr ptr);
+
+        private static string realpath(string path)
+        {
+            IntPtr ptr = realpath(path, IntPtr.Zero);
+            string result = Marshal.PtrToStringAuto(ptr);
+            free(ptr);
+            return result;
+        }
```
rainersigwald, line 125: The .NET convention for this type of thing is to put these into a class named `NativeMethods`: https://docs.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca1060

src/MSBuildLocator/DotNetSdkLocationHelper.cs
```diff
@@ -77,92 +78,122 @@ public static IEnumerable<VisualStudioInstance> GetInstances(string workingDirec
...
-            var lines = new List<string>();
-            try
+            string dotnetPath = null;
+
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
             {
-                process = new Process()
-                { 
-                    StartInfo = new ProcessStartInfo("dotnet", "--info")
+                foreach (string dir in Environment.GetEnvironmentVariable("PATH").Split(';'))
```
rainersigwald, line 133:
```suggestion
                foreach (string dir in Environment.GetEnvironmentVariable("PATH").Split(Path.PathSeparator))
```

This provides semantic meaning and gets rid of semi-magic constants.

src/MSBuildLocator/DotNetSdkLocationHelper.cs
```diff
@@ -77,92 +78,122 @@ public static IEnumerable<VisualStudioInstance> GetInstances(string workingDirec
...
-                // Ensure that we set the DOTNET_CLI_UI_LANGUAGE environment variable to "en-US" before
-                // running 'dotnet --info'. Otherwise, we may get localized results.
-                process.StartInfo.EnvironmentVariables[DOTNET_CLI_UI_LANGUAGE] = "en-US";
-
-                process.OutputDataReceived += (_, e) =>
+                }
+            }
+            else
+            {
+                foreach (string dir in Environment.GetEnvironmentVariable("PATH").Split(':'))
```
rainersigwald, line 144:
```suggestion
                foreach (string dir in Environment.GetEnvironmentVariable("PATH").Split(Path.PathSeparator))
```

src/MSBuildLocator/DotNetSdkLocationHelper.cs
rainersigwald, line 133: Please comment the path search with the reasoning you figured out in consultation with the .NET Runtime folks.

src/MSBuildLocator/DotNetSdkLocationHelper.cs
```diff
@@ -77,92 +78,75 @@ public static IEnumerable<VisualStudioInstance> GetInstances(string workingDirec
...
+            {
+                paths = value;
+            });
 
-                    if (!path.Equals(basePath))
-                        paths.Add(path); 
-                                    
-                    lineSdkIndex++;
-                }
+            if (rc != 0)
```
rainersigwald, line 135: Do we need any means of error reporting here? As is we're silently ignoring failures, which might be reasonable but that isn't fully obvious to me. Justify with a comment?

rainersigwald (commented):
What should the version impact of this be? It's theoretically a same-behavior bugfix but it "feels" bigger than that; I'd probably bump the minor version.

src/MSBuildLocator/DotNetSdkLocationHelper.cs
Forgind: I don't fully understand this (and didn't test it, since I don't know how to introduce an error in hostfxr), but I _think_ errors are automatically printed to stderr:
https://github.com/dotnet/runtime/blob/8e407e32bbb13874a7a8b8651d6d3f1d70574754/src/native/corehost/hostfxr.h#L64

I can add a comment to that effect if you think it reasonable. This part was more aimed at doing something reasonable after writing the error. Note that the docs I looked at all checked the return value but didn't do anything with it other than returning false or whatever.

That said, this made me notice the more important bug: I wasn't resetting rc with the hostfxr_get_available_sdks call, so this part is completely invalid 🙂

Forgind:
> What should the version impact of this be? It's theoretically a same-behavior bugfix but it "feels" bigger than that; I'd probably bump the minor version.

That sounds reasonable to me. If we were pushing to MSBuildLocator regularly, I might push back, but we haven't had a version bump in over a year. It's time. I'm thinking we can push a release version after this goes in and the target framework change goes in. Sound good to you?

Forgind:
Forgot I'd already pushed the version update to the tf PR. Ah, well.

src/MSBuildLocator/DotNetSdkLocationHelper.cs
rainersigwald: Hm. I don't love us writing to stderr accidentally. What I was wondering was whether we should throw an exception if the native call returns anything other than `0`. I think my ideal would be to overwrite that error mechanism, capture it to a string, and throw with that string. What does the SDK resolver do?

Forgind: Basically nothing. They have one replace where they assert it only fails if the directory is null:
https://github.com/dotnet/sdk/blob/a30e465a2e2ea4e2550f319a2dc088daaafe5649/src/Resolvers/Microsoft.DotNet.NativeWrapper/NETCoreSdkResolverNativeWrapper.cs#L25
and they have another case in which they don't check at all:
https://github.com/dotnet/sdk/blob/a30e465a2e2ea4e2550f319a2dc088daaafe5649/src/Resolvers/Microsoft.DotNet.NativeWrapper/NETCoreSdkResolverNativeWrapper.cs#L45

jzabroski: I think, even though those are public methods, they're probably assuming some invariants from the caller.
The right thing to do would be to patch the sdk resolver.

jzabroski: > they have another case in which they don't check at all:
> https://github.com/dotnet/sdk/blob/a30e465a2e2ea4e2550f319a2dc088daaafe5649/src/Resolvers/Microsoft.DotNet.NativeWrapper/NETCoreSdkResolverNativeWrapper.cs#L45

They get the errorCode, and do nothing with it - that should change.

Forgind: You're not wrong; we could do better. On the other hand, I imagine the scenarios in which the difference between "prints to stderr" and "we caught the exception and sent our own error message" really matters are few and far between, noting that the SDK resolver has been used extensively, and that code has peacefully existed for years. I also believe failure here is pretty rare; we aren't doing anything unusual or passing in arguments, so there isn't much that can go wrong.

rainersigwald (approved):

Issue #176, opened 2022-08-14
Version 1.5.3 misses a trailing directory separator path


CamiloTerevinto:
In a project, where the MS Build instance is registered with `MSBuildLocator.RegisterDefaults();` before calling a method that ends up doing:

```
var globalProperties = new Dictionary<string, string>
{
    { "SolutionDir", $"{Path.GetDirectoryName("path")}{Path.DirectorySeparatorChar}" }
};
var projectCollection = new ProjectCollection(globalProperties);
var project = projectCollection.LoadProject("path");
```

Both the above and doing just `new Project("path");` fails with an `InvalidProjectFileException`. 

Looking at the `ProjectCollection`, the difference between Microsoft.Build.Locator 1.4.1 and 1.5.3 is in `ProjectCollection.Toolsets[0]._properties[0]`.

1.4.1: `"MSBuildSDKsPath"="C:\\Program Files\\dotnet\\sdk\\6.0.303\\Sdks"`
1.5.3: `"MSBuildSDKsPath"="C:\\Program Files\\dotnet\\sdk\\6.0.303Sdks"`

This might need to be fixed in the `Microsoft.Build.Evaluation` package instead of here, though.

caaavik-msft:
Just in case it helps anyone else, I got stuck on this for a good chunk of the day as this was the error I was presented which was misleading:
```
Unhandled exception: Microsoft.Build.Exceptions.InvalidProjectFileException: The SDK 'Microsoft.NET.Sdk' specified could not be found.  C:\src\dotnet-performance-tools\test\SampleBenchmarkProject\SampleBenchmarkProject.csproj
   at Microsoft.Build.Shared.ProjectErrorUtilities.ThrowInvalidProject(String errorSubCategoryResourceName, IElementLocation elementLocation, String resourceName, Object[] args)
   at Microsoft.Build.Shared.ProjectErrorUtilities.ThrowInvalidProject[T1](IElementLocation elementLocation, String resourceName, T1 arg0)
   at Microsoft.Build.Evaluation.Evaluator`4.ExpandAndLoadImportsFromUnescapedImportExpressionConditioned(String directoryOfImportingFile, ProjectImportElement importElement, List`1& projects, SdkResult& sdkResult, Boolean throwOnFileNotExistsError)
   at Microsoft.Build.Evaluation.Evaluator`4.ExpandAndLoadImports(String directoryOfImportingFile, ProjectImportElement importElement, SdkResult& sdkResult)
[...]

PR #177, opened 2022-08-16
Bugfix: https://github.com/microsoft/MSBuildLocator/issues/176

bernd5:
closes: #176

dotNetSdkPath may not end with / or \ - we can't use simple concat

### DIFF
--- a/src/MSBuildLocator/MSBuildLocator.cs
+++ b/src/MSBuildLocator/MSBuildLocator.cs
@@ -314,9 +314,24 @@ namespace Microsoft.Build.Locator
             const string MSBuildExtensionsPath = nameof(MSBuildExtensionsPath);
             const string MSBuildSDKsPath = nameof(MSBuildSDKsPath);
 
-            Environment.SetEnvironmentVariable(MSBUILD_EXE_PATH, Path.Combine(dotNetSdkPath, "MSBuild.dll"));
-            Environment.SetEnvironmentVariable(MSBuildExtensionsPath, dotNetSdkPath);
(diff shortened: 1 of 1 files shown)


src/MSBuildLocator/MSBuildLocator.cs
```diff
@@ -314,17 +314,9 @@ private static void ApplyDotNetSdkEnvironmentVariables(string dotNetSdkPath)
...
-                [MSBuildSDKsPath] = dotNetSdkPath + "Sdks"
-            };
-
-            foreach (var kvp in variables)
-            {
-                Environment.SetEnvironmentVariable(kvp.Key, kvp.Value);
-            }
+            Environment.SetEnvironmentVariable(MSBUILD_EXE_PATH, Path.Combine(dotNetSdkPath, "MSBuild.dll"));
+            Environment.SetEnvironmentVariable(MSBuildExtensionsPath, dotNetSdkPath);
+            Environment.SetEnvironmentVariable(MSBuildSDKsPath, Path.Combine(dotNetSdkPath, "Sdks"));
```
caaavik-msft, line 319: I left this comment on the issue, but I'd like to suggest adding sanity checks to ensure that these paths all exist just in case.

bernd5:
Anything to change here?

bernd5:
Please review this PR

Forgind:
Sorry I was a bit slow. This change and #178 seem roughly equivalent, and they're both correct as far as I can tell. That one is a more targeted change, but this one avoids allocating a dictionary unnecessarily. I just happened to see that first, so I merged it. This change looks good, too! Since I already merged that one, I'll close this, but thank you for your contribution!

PR #178, opened 2022-08-19
Use Path.Combine to build environment variables

### LINKED ISSUE #176: Version 1.5.3 misses a trailing directory separator path

Danielku15:
Fixes #176 

As descibed in the issue it seems there are no trailing backslashes in some paths. To ensure proper paths in the environment variables I changed the string concatenation to a safer `Path.Combine`

### DIFF
--- a/src/MSBuildLocator/MSBuildLocator.cs
+++ b/src/MSBuildLocator/MSBuildLocator.cs
@@ -316,9 +316,9 @@ namespace Microsoft.Build.Locator
 
             var variables = new Dictionary<string, string>
             {
-                [MSBUILD_EXE_PATH] = dotNetSdkPath + "MSBuild.dll",
+                [MSBUILD_EXE_PATH] = Path.Combine(dotNetSdkPath, "MSBuild.dll"),
                 [MSBuildExtensionsPath] = dotNetSdkPath,
-                [MSBuildSDKsPath] = dotNetSdkPath + "Sdks"
+                [MSBuildSDKsPath] = Path.Combine(dotNetSdkPath, "Sdks")
             };
 
             foreach (var kvp in variables)


rainersigwald (approved):

Forgind (approved):
I think I'd prefer to have an EnsureTrailingSlash method to adjust the path in each VisualStudioInstance (for core, specifically) then call it before each yield return:
https://github.com/microsoft/MSBuildLocator/blob/master/src/MSBuildLocator/DotNetSdkLocationHelper.cs#L126
and
https://github.com/microsoft/MSBuildLocator/blob/master/src/MSBuildLocator/DotNetSdkLocationHelper.cs#L153
but this is ok.

Forgind:
Thanks @Danielku15!

PR #193, opened 2022-11-10
Update xml doc comments

Forgind:


src/MSBuildLocator/MSBuildLocator.cs
```diff
@@ -15,6 +15,9 @@
 
 namespace Microsoft.Build.Locator
 {
+    /// <summary>
+    /// MSBuildLocator allows the use of publicly visible MSBuild APIs in custom applications. See our documentation at aka.ms/msbuild/locator.
```
rainersigwald, line 19: This doesn't read much like other static classes, for example https://learn.microsoft.com/en-us/dotnet/api/system.gc?view=net-7.0. I like linking to the conceptual doc, though--is it best to do that here like this, or after the XML ingestion, @ghogen?

```suggestion
    /// Allows enumerating installed MSBuild instances and preparing MSBuild APIs for use.
```

src/MSBuildLocator/MSBuildLocator.cs
```diff
@@ -305,7 +308,7 @@ public static void Unregister()
         /// <summary>
         ///     Ensures the proper MSBuild environment variables are populated for DotNet SDK.
         /// </summary>
-        /// <param name="msbuildPath">
+        /// <param name="dotNetSdkPath">
         ///     Path to the directory containing the DotNet SDK.
```
rainersigwald, line 312: We should probably be brand consistent.

```suggestion
        ///     Path to the directory containing the .NET SDK.
```

src/MSBuildLocator/MSBuildLocator.cs
ghogen, line 19:
```suggestion
    /// Allows enumerating installed MSBuild instances and preparing MSBuild APIs for use. See <see href="/visualstudio/msbuild/find-and-use-msbuild-versions">Find and use a version of MSBuild</see>.
```

ghogen: You could use the aka.ms link here, too

ghogen (approved):

rainersigwald (approved):

PR #203, opened 2023-03-16
Use dotnet from current process if possible

Forgind:
If we are running in a dotnet executable, we should use that instead of trying to find it on the PATH.

### DIFF
--- a/src/MSBuildLocator/DotNetSdkLocationHelper.cs
+++ b/src/MSBuildLocator/DotNetSdkLocationHelper.cs
@@ -7,6 +7,7 @@ using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
+using System.Linq;
 using System.Runtime.InteropServices;
 using System.Text.RegularExpressions;
 
@@ -14,9 +15,9 @@ namespace Microsoft.Build.Locator
 {
     internal static class DotNetSdkLocationHelper
     {
-        private static readonly Regex DotNetBasePathRegex = new Regex("Base Path:(.*)$", RegexOptions.Multiline);
         private static readonly Regex VersionRegex = new Regex(@"^(\d+)\.(\d+)\.(\d+)", RegexOptions.Multiline);
-        private static readonly Regex SdkRegex = new Regex(@"(\S+) \[(.*?)]$", RegexOptions.Multiline);
+        private static int rc = 0;
+        private static bool pinnedSdk = false;
 
         public static VisualStudioInstance GetInstance(string dotNetSdkPath)
         {            
@@ -92,6 +93,7 @@ namespace Microsoft.Build.Locator
             bool isWindows = RuntimeInformation.IsOSPlatform(OSPlatform.Windows);
             string exeName = isWindows ? "dotnet.exe" : "dotnet";
             bool returnedSDKs = false;
+            HashSet<string> sdkPaths = new HashSet<string>();
 
             // We will generally find the dotnet exe on the path, but on linux, it is often just a 'dotnet' symlink (possibly even to more symlinks) that we have to resolve
             // to the real dotnet executable.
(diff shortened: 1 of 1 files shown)


ladipro (commented):
This change may break someone. If MSBuildLocator is running under a private runtime, it would now be enumerating only private SDKs in the corresponding directory. Is this really what we want?

ladipro:
.NET runtime is xcopy-able and may exist in many copies on the machine, including copies used only by a specific app. Consider the case where I use MSBuildLocator in an app which runs on its copy of the runtime, like for example on what's under `.dotnet` in our dotnet repos. It's not clear if MSBuildLocator is expected to locate SDK's in this private location (there may be none if it's just runtime without SDK) or in a more conventional location inferred from `PATH`, `DOTNET_ROOT` etc.

What is the motivation for this change?

src/MSBuildLocator/DotNetSdkLocationHelper.cs
```diff
@@ -112,6 +113,31 @@ private static IEnumerable<string> GetDotNetBasePaths(string workingDirectory)
...
+                foreach (string path in GetDotNetBasePaths(workingDirectory, dotnetPath))
+                {
+                    returnedSDKs = true;
+                    yield return path;
+                }
+            }
+
+            if (!returnedSDKs)
+            {
+                throw new InvalidOperationException("Failed to find an appropriate version of .NET Core MSBuild. Call to hostfxr_resolve_sdk2 failed. There may be more details in stderr.");
```
ladipro, line 135: nit: It doesn't look like getting here necessarily implies that `hostfxr_resolve_sdk2` failed. We may have simply found no SDKs.

src/MSBuildLocator/DotNetSdkLocationHelper.cs
```diff
@@ -112,34 +115,110 @@ private static IEnumerable<string> GetDotNetBasePaths(string workingDirectory)
                 dotnetPath = Path.GetDirectoryName(isWindows ? dotnetPath : realpath(dotnetPath) ?? dotnetPath);
             }
 
+            string pathOfPinnedSdk = string.Empty;
+
+            foreach (string path in GetDotNetBasePaths(workingDirectory, dotnetPath))
```
rainersigwald, line 120: What's your thinking on preferring the PATH search to the current executable? I could see that going either way.

src/MSBuildLocator/DotNetSdkLocationHelper.cs
```diff
@@ -112,34 +115,110 @@ private static IEnumerable<string> GetDotNetBasePaths(string workingDirectory)
...
                 if (key == NativeMethods.hostfxr_resolve_sdk2_result_key_t.resolved_sdk_dir)
                 {
                     bestSDK = value;
                 }
             });
 
-            if (rc == 0 && bestSDK != null)
+            if (localRC == 0 && bestSDK != null)
             {
+                pinnedSdk = true;
```
rainersigwald, line 202: What is a "pinned" sdk?

src/MSBuildLocator/DotNetSdkLocationHelper.cs
Forgind: I agree it could go either way...in this case, since I'm adding them all to a HashSet rather than a List (for deduplication), I'm not actually preferring one above the other. I'm just going by version number and mixing them up. (I also see an argument for keeping them separate...but I'm not sure how valuable that is, and I honestly don't think the "from current process" code path will be important all that often.)

src/MSBuildLocator/DotNetSdkLocationHelper.cs
Forgind: That would be like if you have a global.json. Is there a better name that would've made that clear?

YuliiaKovalova:
@baronfel please take a look

YuliiaKovalova:
@Forgind ,could you tell me if I understand it right, that my https://github.com/microsoft/MSBuildLocator/pull/230 covers the implemented functionality?

YuliiaKovalova:
Implemented in scope of https://github.com/microsoft/MSBuildLocator/pull/230

PR #204, opened 2023-03-30
Make Unregister a no-op Fixes #77

### LINKED ISSUE #77: Unregister does not reset IsRegistered to false
If I call `RegisterInstance(...)` and then `Unregister()`, `IsRegistered` continues to return true.

Any code using this to perform conditional registration checks will subsequently not register, and then crash when attempting to resolve an MSBuild assembly.

Forgind:
Unregister currently looks to see if we had previously registered MSBuild, and if so, it undoes our resolver but leaves IsRegistered set to true. This basically means we can't unregister then re-register it, at least not via MSBuildLocator. One fix could be ensuring that IsRegistered is set to false at the end of Unregister, but since we no longer register specific assemblies anyway, making it a no-op seems more reasonable. There's no need to unregister anyway.

### DIFF
--- a/src/MSBuildLocator/MSBuildLocator.cs
+++ b/src/MSBuildLocator/MSBuildLocator.cs
@@ -290,19 +290,6 @@ namespace Microsoft.Build.Locator
         /// </remarks>
         public static void Unregister()
         {
-            if (!IsRegistered)
-            {
-                var error = $"{typeof(MSBuildLocator)}.{nameof(Unregister)} was called, but no MSBuild instance is registered." + Environment.NewLine +
-                            $"Ensure that {nameof(RegisterInstance)}, {nameof(RegisterMSBuildPath)}, or {nameof(RegisterDefaults)} is called before calling this method." + Environment.NewLine +
-                            $"{nameof(IsRegistered)} should be used to determine whether calling {nameof(Unregister)} is a valid operation.";
-                throw new InvalidOperationException(error);
-            }
-
-#if NET46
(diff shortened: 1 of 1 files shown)


src/MSBuildLocator/MSBuildLocator.cs
```diff
@@ -290,19 +290,6 @@ Assembly TryLoadAssembly(AssemblyName assemblyName)
         /// </remarks>
```
rainersigwald, line 290: Change the doc comment too to say "don't call this" please.

rainersigwald (approved):

yaakov-h:
Can it be given ObsoleteAttribute as well, and/or hidden with EditorBrowsable?

Forgind:
> Can it be given ObsoleteAttribute as well, and/or hidden with EditorBrowsable?

Obsolete would throw a warning, so I don't think I can do that, but I think EditorBrowsable is fine! Thanks for the suggestion!

rainersigwald:
Agreed (though we can think about having a less aggressive breaking change policy in this repo than in MSBuild proper).

src/MSBuildLocator/MSBuildLocator.cs
```diff
@@ -8,6 +8,7 @@
 using System.Linq;
 using System.Reflection;
 using System.Text;
+using System.ComponentModel;
```
yaakov-h, line 11: Looking at the existing list, are these normally sorted alphabetically?

Forgind: Fixed, thanks

yaakov-h (approved):

PR #207, opened 2023-04-13
Allow finding Visual Studio installs on Windows on .NET 6+.

Quogu:
This is a very important use-case for me, because the .NET SDKs can't process `.vcxproj` files but Visual Studio can, so being unable to locate Visual Studio build tools at all means I can't run my app on .NET. I think it makes sense to change this given that the readme file explicitly talks about it not being enough to load some projects, so allowing more users to load more projects seems to fit with the project's ethos.

I've tested this on Windows, and it works perfectly for my use-case. Unfortunately I don't have access to other supported .NET platforms, but I'm fairly confident it'll work correctly - `GetDevConsoleInstance` has all the APIs available but will certainly return `null`, and `VisualStudioLocationHelper.GetInstances` will catch a `DllNotFoundException` and continue, so the end result should be the same.

### DIFF
--- a/samples/BuilderApp/BuilderApp.csproj
+++ b/samples/BuilderApp/BuilderApp.csproj
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFrameworks>net472;netcoreapp3.1;net6.0</TargetFrameworks>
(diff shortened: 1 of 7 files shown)


YuliiaKovalova:
The PR seems to be outdated.
Feel free to reopen and address the conflicts if you still need this feature. 
Thank you!

PR #225, opened 2023-08-03
Respect DOTNET_ROOT

Forgind:
Per https://github.com/dotnet/format/issues/1925#issuecomment-1664382538 (see that issue for more context)

### DIFF
--- a/src/MSBuildLocator/DotNetSdkLocationHelper.cs
+++ b/src/MSBuildLocator/DotNetSdkLocationHelper.cs
@@ -78,12 +78,31 @@ namespace Microsoft.Build.Locator
             }
         }
 
+        /// <summary>
+        /// This native method call determines the actual location of path, including
+        /// resolving symbolic links.
+        /// </summary>
         private static string realpath(string path)
         {
             IntPtr ptr = NativeMethods.realpath(path, IntPtr.Zero);
             string result = Marshal.PtrToStringAuto(ptr);
             NativeMethods.free(ptr);
             return result;
+        }
+
+        private static string FindDotnetFromEnvironmentVariable(string environmentVariable, string exeName)
+        {
+            string dotnet_root = Environment.GetEnvironmentVariable(environmentVariable);
+            if (!string.IsNullOrEmpty(dotnet_root))
+            {
+                string fullPathToDotnetFromRoot = Path.Combine(dotnet_root, exeName);
+                if (File.Exists(fullPathToDotnetFromRoot))
+                {
+                    return realpath(fullPathToDotnetFromRoot) ?? fullPathToDotnetFromRoot;
+                }
+            }
+
+            return null;
         }
 
         private static IEnumerable<string> GetDotNetBasePaths(string workingDirectory)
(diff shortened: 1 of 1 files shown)


src/MSBuildLocator/DotNetSdkLocationHelper.cs
```diff
@@ -92,18 +92,37 @@ private static IEnumerable<string> GetDotNetBasePaths(string workingDirectory)
...
-            // This can be done using the nethost library. We didn't do this previously, so I did not implement this extension.
-            foreach (string dir in Environment.GetEnvironmentVariable("PATH").Split(Path.PathSeparator))
-            {
-                string filePath = Path.Combine(dir, exeName);
-                if (File.Exists(Path.Combine(dir, exeName)))
+            // First check for the DOTNET_ROOT environment variable, as it's often there as with, for example, dotnet format.
+            string dotnet_root = Environment.GetEnvironmentVariable("DOTNET_ROOT");
+            if (!string.IsNullOrEmpty(dotnet_root))
+            {
+                // DOTNET_ROOT can be a path to dotnet OR a path to the folder containing dotnet.
```
NikolaMilosavljevic, line 99: I think this can only be a folder.

https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-environment-variables

Also:
https://github.com/dotnet/runtime/blob/main/docs/design/features/host-components.md
https://github.com/dotnet/runtime/blob/10d9e729f9ece24f89d73a5c6df18371c8db9360/src/installer/tests/HostActivation.Tests/MultiArchInstallLocation.cs#L119

YuliiaKovalova (approved):

rainersigwald:
`DOTNET_ROOT` seems analogous to `DiscoveryOptions.DeveloperConsole`. Should we instead extend that concept and codepath to respect it (as well as the VS environment variables)?

Forgind:
> `DOTNET_ROOT` seems analogous to `DiscoveryOptions.DeveloperConsole`. Should we instead extend that concept and codepath to respect it (as well as the VS environment variables)?

There are definitely some similarities. I don't think there's a lot of overlap, though, which would make it tricky to try to reuse too much code—DiscoveryOptions.DeveloperConsole is specifically intended to be for NetFx users of VS developer command prompts, a very specific scenario. Of note, the environment variables it checks for there (VSINSTALLDIR, VSCMD_VER, and VisualStudioVersion) are rather VS-specific and not relevant to the DotNetSdkLocationHelper.

So I guess I'm not entirely clear on what concrete changes you're proposing? Though as I was rooting (hehe) around, I did notice there's also a DOTNET_ROOT(x86), so I should probably take that into account as well even though no one (to my knowledge) has specifically requested that yet.

src/MSBuildLocator/DotNetSdkLocationHelper.cs
```diff
@@ -92,27 +92,55 @@ private static IEnumerable<string> GetDotNetBasePaths(string workingDirectory)
...
                 {
-                    if (!isWindows)
+                    dotnetPath = fullPathToDotnetFromRoot;
+                }
+            }
+
+            // Second, check for the DOTNET_ROOT(x86) environment variable, as it can be there, too.
+            if (dotnetPath is null)
+            {
+                string dotnet_root_x86 = Environment.GetEnvironmentVariable("DOTNET_ROOT(x86)");
```
YuliiaKovalova, line 109: Can't it be moved to a separate method as code starting from line 96?

YuliiaKovalova:
@baronfel , @rainersigwald any blockers for this one?

src/MSBuildLocator/DotNetSdkLocationHelper.cs
```diff
@@ -92,27 +92,55 @@ private static IEnumerable<string> GetDotNetBasePaths(string workingDirectory)
...
+                // This can be done using the nethost library. We didn't do this previously, so I did not implement this extension.
+                foreach (string dir in Environment.GetEnvironmentVariable("PATH").Split(Path.PathSeparator))
+                {
+                    string filePath = Path.Combine(dir, exeName);
+                    if (File.Exists(filePath))
+                    {
+                        if (!isWindows)
                         {
-                            continue;
+                            filePath = realpath(filePath) ?? filePath;
```
JanKrivanek, line 134: Comment would be helpful here ('unwrapping symlinks' or so)

JanKrivanek (approved):
Looks good

src/MSBuildLocator/DotNetSdkLocationHelper.cs
```diff
@@ -92,27 +92,55 @@ private static IEnumerable<string> GetDotNetBasePaths(string workingDirectory)
...
+                string fullPathToDotnetFromRoot = Path.Combine(dotnet_root, exeName);
+                if (File.Exists(fullPathToDotnetFromRoot))
                 {
-                    if (!isWindows)
+                    dotnetPath = fullPathToDotnetFromRoot;
+                }
+            }
+
+            // Second, check for the DOTNET_ROOT(x86) environment variable, as it can be there, too.
+            if (dotnetPath is null)
```
rainersigwald, line 107: Should this be filtered to only if we are currently in an x86 process? It feels weird to respect `DOTNET_ROOT(x86)` in an amd64/x64 process.

src/MSBuildLocator/DotNetSdkLocationHelper.cs
```diff
@@ -92,27 +92,55 @@ private static IEnumerable<string> GetDotNetBasePaths(string workingDirectory)
...
-                if (File.Exists(filePath))
+            // First check for the DOTNET_ROOT environment variable, as it's often there as with, for example, dotnet format.
+            string dotnet_root = Environment.GetEnvironmentVariable("DOTNET_ROOT");
+            if (!string.IsNullOrEmpty(dotnet_root))
+            {
+                string fullPathToDotnetFromRoot = Path.Combine(dotnet_root, exeName);
+                if (File.Exists(fullPathToDotnetFromRoot))
                 {
-                    if (!isWindows)
+                    dotnetPath = fullPathToDotnetFromRoot;
```
rainersigwald, line 102: Does this need `realpath`?

Issue #210, opened 2023-04-18
MSBuildLocator.QueryVisualStudioInstances Fails on Mac and Linux from Version 1.5.3


jamesHargreaves12:
Starting from version [1.5.3](https://www.nuget.org/packages/Microsoft.Build.Locator/1.5.3), the MSBuildLocator.QueryVisualStudioInstances method fails to execute in Mac and Linux environments. The error message is as follows:
```
One or more errors occurred. (Unable to load shared library 'hostfxr' or one of its dependencies. In order to help diagnose loading problems, consider setting the DYLD_PRINT_LIBRARIES environment variable: dlopen(libhostfxr, 1): image not found)
  Inner exception: Unable to load shared library 'hostfxr' or one of its dependencies. In order to help diagnose loading problems, consider setting the DYLD_PRINT_LIBRARIES environment variable: dlopen(libhostfxr, 1): image not found
System.AggregateException: One or more errors occurred. (Unable to load shared library 'hostfxr' or one of its dependencies. In order to help diagnose loading problems, consider setting the DYLD_PRINT_LIBRARIES environment variable: dlopen(libhostfxr, 1): image not found)
 ---> System.DllNotFoundException: Unable to load shared library 'hostfxr' or one of its dependencies. In order to help diagnose loading problems, consider setting the DYLD_PRINT_LIBRARIES environment variable: dlopen(libhostfxr, 1): image not found
   at Microsoft.Build.Locator.NativeMethods.hostfxr_resolve_sdk2(String exe_dir, String working_dir, hostfxr_resolve_sdk2_flags_t flags, hostfxr_resolve_sdk2_result_fn result)
   at Microsoft.Build.Locator.DotNetSdkLocationHelper.GetDotNetBasePaths(String workingDirectory)+MoveNext()
   at Microsoft.Build.Locator.DotNetSdkLocationHelper.GetInstances(String workingDirectory)+MoveNext()
   at Microsoft.Build.Locator.MSBuildLocator.GetInstances(VisualStudioInstanceQueryOptions options)+MoveNext()
```
It's worth noting that the QueryInstancesTests.DefaultInstanceTest unit test fails with the same issue on my mac.

The underlying cause of this issue is the `DllImport("hostfxr"...)` fails. It cannot find the hostfxr library without additional information about its location. There are similar issues previously reported on dotnet previews, such as https://github.com/dotnet/runtime/issues/54965. The dotnet team resolved these issues by:
[...]

Issue #200, opened 2023-03-13
Doesn't work on Apple M2


bartofzo:
Found out about this trying to build a docfx project on my MBP:
https://github.com/dotnet/docfx/issues/8502#issue-1620402347

When running a .NET core console app and calling MSBuildLocator.RegisterDefaults()
I get the following exception:

Unhandled exception. System.InvalidOperationException: Failed to find an appropriate version of .NET Core MSBuild. Call to hostfxr_resolve_sdk2 failed. There may be more details in stderr.
at Microsoft.Build.Locator.DotNetSdkLocationHelper.GetDotNetBasePaths(String workingDirectory)+MoveNext()
at Microsoft.Build.Locator.DotNetSdkLocationHelper.GetInstances(String workingDirectory)+MoveNext()
at Microsoft.Build.Locator.MSBuildLocator.GetInstances(VisualStudioInstanceQueryOptions options)+MoveNext()
at System.Linq.Enumerable.TryGetFirst[TSource](IEnumerable1 source, Boolean& found) at System.Linq.Enumerable.FirstOrDefault[TSource](IEnumerable1 source)
at Microsoft.Build.Locator.MSBuildLocator.RegisterDefaults()
at Program.

OS: MacOS ventura 13.2.1, Apple M2 Pro
Never had this problem on an Intel machine so I suspect it is due to that. Any fix or workaround available?

Forgind:
Can you check if your dotnet executable is on your PATH? And if it's a symlink, can you make sure its target is the right dotnet?

bartofzo:
I think it is, but I'm no expert at these kinds of things. If i run dotnet --info from a terminal I get the following information:

`Runtime Environment:
 OS Name:     Mac OS X
 OS Version:  13.2
 OS Platform: Darwin
 RID:         osx.13-arm64
 Base Path:   /opt/homebrew/Cellar/dotnet/7.0.100/libexec/sdk/7.0.100/

Host:
  Version:      7.0.0
  Architecture: arm64
  Commit:       d099f075e4

.NET SDKs installed:
  7.0.100 [/opt/homebrew/Cellar/dotnet/7.0.100/libexec/sdk]

.NET runtimes installed:
  Microsoft.AspNetCore.App 7.0.0 [/opt/homebrew/Cellar/dotnet/7.0.100/libexec/shared/Microsoft.AspNetCore.App]
[...]

Issue #181, opened 2022-09-03
Regression in v1.5.3 -  The SDK 'Microsoft.NET.Sdk' specified could not be found. 


kamronbatman:
I updated our dependency in this project from 1.4.1 to 1.5.3: https://github.com/modernuo/SerializationGenerator/tree/main/ModernUO.Serialization.SchemaGenerator and now I am getting the error `The SDK 'Microsoft.NET.Sdk' specified could not be found.`

I have the VS2022 and 6.0.400 SDK installed. I do not have .NET 7 preview installed.

EDIT: Now I have a weird issue. I decided to not update this package and leave it at 1.4.1. But I did update WorkSpaces to 4.3.0. And it seems to also be broken, but only when I use the nuget package I built, not in the IDE (either VS or Rider).

This is super weird and a new problem. I am going to try and revert back to workspaces v4.2.0 and see if that fixes it.

EDIT 2: It looks like reverting to msbuildlocator 1.4.1 and workspaces 4.2.0 fixed it. So I am not sure if this is even an issue with this package. I would love some advisement since I am a little lost tracking it all.

Also to note: MSBuildLocator 1.5.3 and Workspaces 4.2.0/4.3.0 do not work. And with Workspaces 4.3.0, MacOS 11 gets a hostfx native image error while trying to find the SDK, which I think is a separate issue, but relevant to my testing since I test on a variety of operating systems.

Forgind:
Failing to find the SDK sounds like either an MSBuild bug or an SDK bug, as MSBuild owns the general process of SDK resolution, and the SDK owns the SDK resolver that should work in this case. I think Roslyn owns the MSBuildWorkspace, though I'm not sure about that.

The hostfxr bug is very plausibly an MSBuildLocator issue, though. Do you have a path to your dotnet executable on the PATH? And is there another (earlier) folder on your PATH that has either a bad executable or a symlink that looks reasonable but actually doesn't point to a file that exists? One of my recent (unmerged) PRs is trying to fix the latter problem.

Forgind:
I'd also check out https://github.com/microsoft/MSBuildLocator/issues/68#issuecomment-1384609721, as that might be the issue.

kamronbatman:
I have wiped my MacOS since this issue. I'll do a clean setup with .net 7 and bump dependencies and see if I get the issue. I think #68 might be related but for a different reason since this is OSX.

YuliiaKovalova:
The issue is fixed. Pending release.

PR #230, opened 2023-08-22
Fix resolving hostfxr

YuliiaKovalova:
Fixes:
https://github.com/dotnet/msbuild/issues/9038
https://github.com/microsoft/MSBuildLocator/issues/210
https://github.com/microsoft/MSBuildLocator/issues/200
https://github.com/microsoft/MSBuildLocator/issues/181

This PR solves 2 problems:
1. Dotnet path resolution
2. Loading hostfx.(dylib|so) for MAC OS and Linux

In order to resolve dotnet path, we probe different sources:
Case: MSBuild.Locator is called **from dotnet.exe,** check DOTNET_ROOT env var -> current process path
Case: MSBuild.Locator is called **in apphost,** check DOTNET_ROOT env var -> DOTNET_HOST_PATH -> DOTNET_MSBUILD_SDK_RESOLVER_CLI_DIR -> PATH
If the path can't be extracted, an exception is thrown.

When DotnetPath is resolved, we create a subscription on ResolvingUnmanagedDll event when hostfx is requested, if it isn't not loaded in context (the possible reason is described here https://github.com/dotnet/runtime/issues/83774).
The path to hostfxr.dylib|so is built based on the pattern <dotnet_path>/host/fxr/<version>/libhostfxr.(dylib|so)

### DIFF
--- a/src/MSBuildLocator/DotNetSdkLocationHelper.cs
+++ b/src/MSBuildLocator/DotNetSdkLocationHelper.cs
@@ -3,12 +3,14 @@
 
 #if NETCOREAPP
 
+using NuGet.Versioning;
 using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
 using System.Reflection;
 using System.Runtime.InteropServices;
+using System.Runtime.Loader;
 using System.Text.RegularExpressions;
 
 #nullable enable
@@ -20,29 +22,35 @@ namespace Microsoft.Build.Locator
         private static readonly Regex VersionRegex = new Regex(@"^(\d+)\.(\d+)\.(\d+)", RegexOptions.Multiline);
         private static readonly bool IsWindows = RuntimeInformation.IsOSPlatform(OSPlatform.Windows);
         private static readonly string ExeName = IsWindows ? "dotnet.exe" : "dotnet";
-        private static readonly string DotnetPath = ResolveDotnetPath();
-
-        static DotNetSdkLocationHelper() => LoadHostFxr();
+        private static readonly Lazy<string> DotnetPath = new(() => ResolveDotnetPath());
 
         public static VisualStudioInstance? GetInstance(string dotNetSdkPath)
         {
             if (string.IsNullOrWhiteSpace(dotNetSdkPath) || !File.Exists(Path.Combine(dotNetSdkPath, "Microsoft.Build.dll")))
+            {
                 return null;
+            }
 
             string versionPath = Path.Combine(dotNetSdkPath, ".version");
             if (!File.Exists(versionPath))
+            {
                 return null;
+            }
 
             // Preview versions contain a hyphen after the numeric part of the version. Version.TryParse doesn't accept that.
             Match versionMatch = VersionRegex.Match(File.ReadAllText(versionPath));
 
             if (!versionMatch.Success)
+            {
                 return null;
+            }
 
             if (!int.TryParse(versionMatch.Groups[1].Value, out int major) ||
                 !int.TryParse(versionMatch.Groups[2].Value, out int minor) ||
                 !int.TryParse(versionMatch.Groups[3].Value, out int patch))
+            {
                 return null;
+            }   
 
             // Components of the SDK often have dependencies on the runtime they shipped with, including that several tasks that shipped
             // in the .NET 5 SDK rely on the .NET 5.0 runtime. Assuming the runtime that shipped with a particular SDK has the same version,
@@ -50,7 +58,9 @@ namespace Microsoft.Build.Locator
             // to always work but should work for now.
             if (major > Environment.Version.Major ||
                 (major == Environment.Version.Major && minor > Environment.Version.Minor))
+            {
                 return null;
+            }
 
             return new VisualStudioInstance(
                 name: ".NET Core SDK",
@@ -65,87 +75,91 @@ namespace Microsoft.Build.Locator
             {
                 var dotnetSdk = GetInstance(basePath);
                 if (dotnetSdk != null)
+                {
                     yield return dotnetSdk;
+                }
             }
         }
 
-        public static IEnumerable<string> GetDotNetBasePaths(string workingDirectory)
+        private static IEnumerable<string> GetDotNetBasePaths(string workingDirectory)
         {
-            string? bestSDK = GetSdkFromGlobalSettings(workingDirectory);
-            if (!string.IsNullOrEmpty(bestSDK))
-                yield return bestSDK;
-
-            string[] dotnetPaths = GetAllAvailableSDKs();
-            // We want to return the newest SDKs first, however, so iterate over the list in reverse order.
-            // If basePath is disqualified because it was later
-            // than the runtime version, this ensures that RegisterDefaults will return the latest valid
-            // SDK instead of the earliest installed.
-            for (int i = dotnetPaths.Length - 1; i >= 0; i--)
+            try
             {
-                if (dotnetPaths[i] != bestSDK)
-                    yield return dotnetPaths[i];
+                AddUnmanagedDllResolver();
+
+                string? bestSDK = GetSdkFromGlobalSettings(workingDirectory);
+                if (!string.IsNullOrEmpty(bestSDK))
+                {
+                    yield return bestSDK;
+                }
+
+                string[] dotnetPaths = GetAllAvailableSDKs();
+                // We want to return the newest SDKs first, however, so iterate over the list in reverse order.
+                // If basePath is disqualified because it was later
+                // than the runtime version, this ensures that RegisterDefaults will return the latest valid
+                // SDK instead of the earliest installed.
+                for (int i = dotnetPaths.Length - 1; i >= 0; i--)
+                {
+                    if (dotnetPaths[i] != bestSDK)
+                    {
+                        yield return dotnetPaths[i];
+                    }
+                }
+            }
+            finally
+            {
+                RemoveUnmanagedDllResolver();
             }
         }
 
-        private static void LoadHostFxr()
+        private static void AddUnmanagedDllResolver() => ModifyUnmanagedDllResolver(loadContext => loadContext.ResolvingUnmanagedDll += HostFxrResolver);
+
+        private static void RemoveUnmanagedDllResolver() => ModifyUnmanagedDllResolver(loadContext => loadContext.ResolvingUnmanagedDll -= HostFxrResolver);
+
+        private static void ModifyUnmanagedDllResolver(Action<AssemblyLoadContext> resolverAction)
         {
-            var isOSX = RuntimeInformation.IsOSPlatform(OSPlatform.OSX);
-            if (isOSX)
+            if (!IsWindows)
             {
-                var loadContext = System.Runtime.Loader.AssemblyLoadContext.GetLoadContext(Assembly.GetExecutingAssembly());
+                var loadContext = AssemblyLoadContext.GetLoadContext(Assembly.GetExecutingAssembly());
                 if (loadContext != null)
-                    loadContext.ResolvingUnmanagedDll += HostFxrResolver;
+                {
+                    resolverAction(loadContext);
+                }
             }
         }
 
         private static IntPtr HostFxrResolver(Assembly assembly, string libraryName)
         {
-            var hostFxrLibName = "libhostfxr.dylib";
+            var hostFxrLibName = "libhostfxr";
+            var libExtention = RuntimeInformation.IsOSPlatform(OSPlatform.OSX) ? "dylib" : "so";
 
-            if (string.Equals(hostFxrLibName, libraryName, StringComparison.Ordinal))
+            if (!hostFxrLibName.Equals(libraryName, StringComparison.OrdinalIgnoreCase))
+            {
                 return IntPtr.Zero;
+            }
 
-            var hostFxrRoot = Path.Combine(DotnetPath, "host", "fxr");
+            var hostFxrRoot = Path.Combine(DotnetPath.Value, "host", "fxr");
             if (Directory.Exists(hostFxrRoot))
             {
                 // Agreed to load hostfxr from the highest version
                 var hostFxrAssemblyDirectory = Directory.GetDirectories(hostFxrRoot)
-                    .OrderByDescending(d => d)
-                    .FirstOrDefault();
+                            .Select(str => NuGetVersion.Parse(str))
+                            .Max();
 
-                if (hostFxrAssemblyDirectory != null)
+                if (hostFxrAssemblyDirectory != null && !string.IsNullOrEmpty(hostFxrAssemblyDirectory.OriginalVersion))
                 {
-                    var hostfxrAssembly = Directory.GetFiles(hostFxrAssemblyDirectory)
-                        .Where(filePath => filePath.Contains(hostFxrLibName, StringComparison.Ordinal))
+                    var hostfxrAssembly = Directory.GetFiles(hostFxrAssemblyDirectory.OriginalVersion)
+                        .Where(filePath => filePath.Equals(Path.Combine(hostFxrLibName, libExtention), StringComparison.OrdinalIgnoreCase))
                         .FirstOrDefault();
 
                     if (hostfxrAssembly != null)
+                    {
                         return NativeLibrary.TryLoad(hostfxrAssembly, out var handle) ? handle : IntPtr.Zero;
+                    }
                 }
             }
 
             return IntPtr.Zero;
-        }
-
-        private static string? FindDotnetFromEnvironmentVariable(string environmentVariable, string exeName)
-        {
-            string? dotnet_root = Environment.GetEnvironmentVariable(environmentVariable);
-            if (!string.IsNullOrEmpty(dotnet_root))
-            {
-                string fullPathToDotnetFromRoot = Path.Combine(dotnet_root, exeName);
-                if (File.Exists(fullPathToDotnetFromRoot))
-                {
-                    if (!IsWindows)
-                    {
-                        fullPathToDotnetFromRoot = realpath(fullPathToDotnetFromRoot) ?? fullPathToDotnetFromRoot;
-                        return File.Exists(fullPathToDotnetFromRoot) ? Path.GetDirectoryName(fullPathToDotnetFromRoot) : null;
-                    }
-
-                    return dotnet_root;
-                }
-            }
-
-            return null;
         }
 
         private static string SdkResolutionExceptionMessage(string methodName) => $"Failed to find all versions of .NET Core MSBuild. Call to {methodName}. There may be more details in stderr.";
@@ -157,14 +171,18 @@ namespace Microsoft.Build.Locator
         private static string? GetSdkFromGlobalSettings(string workingDirectory)
         {
             string? resolvedSdk = null;
-            int rc = NativeMethods.hostfxr_resolve_sdk2(exe_dir: DotnetPath, working_dir: workingDirectory, flags: 0, result: (key, value) =>
+            int rc = NativeMethods.hostfxr_resolve_sdk2(exe_dir: DotnetPath.Value, working_dir: workingDirectory, flags: 0, result: (key, value) =>
             {
                 if (key == NativeMethods.hostfxr_resolve_sdk2_result_key_t.resolved_sdk_dir)
+                {
                     resolvedSdk = value;
+                }
             });
 
             if (rc != 0)
+            {
                 throw new InvalidOperationException(SdkResolutionExceptionMessage(nameof(NativeMethods.hostfxr_resolve_sdk2)));
+            }
 
             return resolvedSdk;
         }
@@ -180,7 +198,9 @@ namespace Microsoft.Build.Locator
                     && Path.GetFileName(dotnetExePath).Equals(ExeName, StringComparison.InvariantCultureIgnoreCase);
                 
                 if (isRunFromDotnetExecutable)
+                {
                     dotnetPath = Path.GetDirectoryName(dotnetExePath);
+                }
                 else
                 {
                     dotnetPath = FindDotnetPathFromEnvVariable("DOTNET_HOST_PATH") 
@@ -190,7 +210,9 @@ namespace Microsoft.Build.Locator
             }
 
             if (string.IsNullOrEmpty(dotnetPath))
+            {
                 throw new InvalidOperationException("Could not find the dotnet executable. Is it set on the DOTNET_ROOT?");
+            }
 
             return dotnetPath;
         }
@@ -204,27 +226,7 @@ namespace Microsoft.Build.Locator
             return dotnetPath;
         }
 
-        private static string? GetDotnetPathFromHOST()
-        {
-            var dotnetPath = FindDotnetPathFromEnvVariable("DOTNET_HOST_PATH");
-            if (dotnetPath == null)
-            {
-                dotnetPath ??= FindDotnetPathFromEnvVariable("DOTNET_MSBUILD_SDK_RESOLVER_CLI_DIR");
-            }
-
-            return dotnetPath;
-        }
-
-        private static string? GetCurrentProcessPath()
-        {
-            string? processPath = null;
-#if NET6_0_OR_GREATER
-            processPath = Environment.ProcessPath;
-#else
-            processPath = Process.GetCurrentProcess().MainModule.FileName;
-#endif
-            return processPath;
-        }
+        private static string? GetCurrentProcessPath() => Environment.ProcessPath;
 
         private static string? GetDotnetPathFromPATH()
         {
@@ -239,10 +241,11 @@ namespace Microsoft.Build.Locator
             foreach (string dir in paths)
             {
                 string? filePath = ValidatePath(dir);
-                if (string.IsNullOrEmpty(filePath))
-                    continue;
-
-                dotnetPath = dir;
+                if (!string.IsNullOrEmpty(filePath))
+                {
+                    dotnetPath = filePath;
+                    break;
+                }
             }
 
             return dotnetPath;
@@ -254,11 +257,13 @@ namespace Microsoft.Build.Locator
         private static string[] GetAllAvailableSDKs()
         {
             string[]? resolvedPaths = null;
-            int rc = NativeMethods.hostfxr_get_available_sdks(exe_dir: DotnetPath, result: (key, value) => resolvedPaths = value);
+            int rc = NativeMethods.hostfxr_get_available_sdks(exe_dir: DotnetPath.Value, result: (key, value) => resolvedPaths = value);
 
             // Errors are automatically printed to stderr. We should not continue to try to output anything if we failed.
             if (rc != 0)
+            {
                 throw new InvalidOperationException(SdkResolutionExceptionMessage(nameof(NativeMethods.hostfxr_get_available_sdks)));
+            }
 
             return resolvedPaths ?? Array.Empty<string>();
         }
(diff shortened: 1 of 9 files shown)


YuliiaKovalova:
@baronfel please take a look

src/MSBuildLocator/DotNetSdkLocationHelper.cs
```diff
@@ -78,124 +67,251 @@ public static IEnumerable<VisualStudioInstance> GetInstances(string workingDirec
...
-                    if (File.Exists(filePath))
+                    // Agreed to load hostfxr from the highest version
+                    var hostFxrAssemblyDirectory = Directory.GetDirectories(hostFxrRoot)
+                        .OrderByDescending(d => d)
+                        .FirstOrDefault();
+
+                    if (hostFxrAssemblyDirectory != null)
                     {
-                        if (!isWindows)
+                        Regex regex = new Regex("(lib)?hostfxr.dylib", RegexOptions.IgnoreCase);
```
Forgind, line 194: We should probably not ignore case on mac.

Also, using a regex feels like overkill here. If there are only two possibilities, you could just check both.

src/MSBuildLocator/DotNetSdkLocationHelper.cs
```diff
@@ -78,124 +67,251 @@ public static IEnumerable<VisualStudioInstance> GetInstances(string workingDirec
...
+
+                        if (matchingAssembly != null)
                         {
-                            filePath = realpath(filePath) ?? filePath;
-                            if (!File.Exists(filePath))
-                            {
-                                continue;
-                            }
-                        }
+                            var isHostFxrLoaded = AppDomain.CurrentDomain.GetAssemblies().Any(assembly =>
```
Forgind, line 202: Should we also try to load hostfxr first? There may be a resolver for it that hasn't fired yet because no one has yet asked for it.

src/MSBuildLocator/DotNetSdkLocationHelper.cs
```diff
@@ -78,124 +67,251 @@ public static IEnumerable<VisualStudioInstance> GetInstances(string workingDirec
...
+        {
+            // 32-bit architecture has (x86) suffix
+            string envVarName = (IntPtr.Size == 4) ? "DOTNET_ROOT(x86)" : "DOTNET_ROOT";
+            var dotnetPath = FindDotnetPathFromEnvVariable(envVarName);
+            
+            return dotnetPath;
+        }
 
-            if (dotnetPath is null)
+        private static string? GetDotnetPathFromDefaultPaths()
```
Forgind, line 222: This doesn't seem to be doing what the name suggests? I'd expected it to look in specific predefined directories.

src/MSBuildLocator/DotNetSdkLocationHelper.cs
```diff
@@ -78,124 +67,251 @@ public static IEnumerable<VisualStudioInstance> GetInstances(string workingDirec
...
+        {
+            string? dotnetPath = GetDotnetPathFromROOT();
+
+            if (string.IsNullOrEmpty(dotnetPath))
             {
-                // 64-bit architecture
-                dotnetPath ??= FindDotnetFromEnvironmentVariable("DOTNET_ROOT", exeName);
+                string? dotnetExePath = GetCurrentProcessPath();
+                var isRunFromDotnetExecutable = !string.IsNullOrEmpty(dotnetExePath) 
+                    && Path.GetFileNameWithoutExtension(dotnetExePath).Equals(ExeName, StringComparison.InvariantCultureIgnoreCase);
```
Forgind, line 160: without extension? Doesn't the exename include the extension?

src/MSBuildLocator/DotNetSdkLocationHelper.cs
```diff
@@ -78,124 +67,251 @@ public static IEnumerable<VisualStudioInstance> GetInstances(string workingDirec
...
+            if (isOSX)
+            {
+                var loadContext = System.Runtime.Loader.AssemblyLoadContext.GetLoadContext(Assembly.GetExecutingAssembly());
+                if (loadContext != null)
+                    loadContext.ResolvingUnmanagedDll += HostFxrResolver;
+            }
         }
 
-        private static string FindDotnetFromEnvironmentVariable(string environmentVariable, string exeName)
+        private static IntPtr HostFxrResolver(Assembly assembly, string libraryName)
```
Forgind, line 100: Why do you have separate LoadHostFxr(string) and LoadHostFxr() methods? From what I see, the former is never used.

src/MSBuildLocator/DotNetSdkLocationHelper.cs
```diff
@@ -78,124 +67,251 @@ public static IEnumerable<VisualStudioInstance> GetInstances(string workingDirec
...
+                throw new InvalidOperationException(SdkResolutionExceptionMessage(nameof(NativeMethods.hostfxr_resolve_sdk2)));
+
+            return resolvedSdk;
+        }
+
+        private static string ResolveDotnetPath()
+        {
+            string? dotnetPath = GetDotnetPathFromROOT();
+
+            if (string.IsNullOrEmpty(dotnetPath))
```
Forgind, line 156: Most of these helper methods return null on failure, correct? Should be able to have a series of:

dotnetPath ??= thing();

-type calls. Should be easier to understand.

Forgind (commented):
I'm not necessarily opposed to it, but this feels like a lot of changes for not very many impactful changes. I generally prefer more targeted changes.

Also, can you separate the framework changes into a separate PR? That feels meaningfully distinct.

ladipro:
> Also, can you separate the framework changes into a separate PR? That feels meaningfully distinct.

+1 on a separate PR for the target framework change.

Also, please describe the changes in the initial comment. Ideally everyone would be able to understand _how_ the PR is fixing the issue by reading the description.

YuliiaKovalova:
> > Also, can you separate the framework changes into a separate PR? That feels meaningfully distinct.
> 
> +1 on a separate PR for the target framework change.
> 
> Also, please describe the changes in the initial comment. Ideally everyone would be able to understand _how_ the PR is fixing the issue by reading the description.

I have moved this change to a separate PR:
https://github.com/microsoft/MSBuildLocator/pull/231 please take a look.

and the second part of comment is relevant - I will describe my steps in detail.

src/MSBuildLocator/DotNetSdkLocationHelper.cs
YuliiaKovalova: Good catch! Thanks :)

src/MSBuildLocator/DotNetSdkLocationHelper.cs
YuliiaKovalova: removed. I will configure code styles in a separate PR to avoid these dangling things.

src/MSBuildLocator/DotNetSdkLocationHelper.cs
YuliiaKovalova: I applied your recommendation in the else statement - looks great.

Here I would leave branching as it is because we have different probing logic for MsBuild.Location run from dotnet.exe or apphost.

src/MSBuildLocator/DotNetSdkLocationHelper.cs
```diff
@@ -79,134 +69,206 @@ public static IEnumerable<VisualStudioInstance> GetInstances(string workingDirec
...
-            {
-                // 32-bit architecture
-                dotnetPath ??= FindDotnetFromEnvironmentVariable("DOTNET_ROOT(x86)", exeName);
-            }
-            else if (IntPtr.Size == 8)
-            {
-                // 64-bit architecture
-                dotnetPath ??= FindDotnetFromEnvironmentVariable("DOTNET_ROOT", exeName);
-            }
+            if (!hostFxrLibName.Contains(libraryName, StringComparison.Ordinal))
```
ladipro, line 105: Why contains instead of equals? Should we be stricter and check that `libraryName` is `libhostfxr`, case insensitive?

src/MSBuildLocator/DotNetSdkLocationHelper.cs
```diff
@@ -79,134 +69,206 @@ public static IEnumerable<VisualStudioInstance> GetInstances(string workingDirec
...
+            string envVarName = (IntPtr.Size == 4) ? "DOTNET_ROOT(x86)" : "DOTNET_ROOT";
+            var dotnetPath = FindDotnetPathFromEnvVariable(envVarName);
+            
+            return dotnetPath;
+        }
+
+        private static string? GetCurrentProcessPath()
+        {
+            string? processPath = null;
+#if NET6_0_OR_GREATER
```
ladipro, line 189: nit: Isn't this always true? Is there a reason for compiling this file against .NET Framework?

src/MSBuildLocator/DotNetSdkLocationHelper.cs
```diff
@@ -5,62 +5,52 @@
...
+        private static readonly string ExeName = IsWindows ? "dotnet.exe" : "dotnet";
+        private static readonly string DotnetPath = ResolveDotnetPath();
 
-        public static VisualStudioInstance GetInstance(string dotNetSdkPath)
-        {            
-            if (string.IsNullOrWhiteSpace(dotNetSdkPath))
-            {
-                return null;
-            }
+        static DotNetSdkLocationHelper() => LoadHostFxr();
```
ladipro, line 25: Why do these methods run as part of static/type initialization? It would be better to run them when actually needed, one reason is exception handling where we would prefer determinism as to when exceptions are thrown.

YuliiaKovalova: In order to run LoadHostFxr(), we need to Resolve dotnet path. Nathan pointed out in his comment that we could potentially fix other hostfxr-related problems
https://github.com/microsoft/MSBuildLocator/pull/230#discussion_r1302149135

YuliiaKovalova: If you think it's not our case, I can move it back in the method invocation.

src/MSBuildLocator/DotNetSdkLocationHelper.cs
YuliiaKovalova: done

src/MSBuildLocator/DotNetSdkLocationHelper.cs
ladipro: I'm not sure I understand the comment. This is an internal class and we control the order of method invocations. Doing this work in the static constructor would be bad practice.

src/MSBuildLocator/DotNetSdkLocationHelper.cs
ladipro: I thought that #231 retargeted the project to net6.0. Is it really possible for this code to run on net5.0 or lower?

YuliiaKovalova: Oh, yes. Fixed

src/MSBuildLocator/DotNetSdkLocationHelper.cs
```diff
@@ -75,138 +73,227 @@ public static IEnumerable<VisualStudioInstance> GetInstances(string workingDirec
...
+                string? filePath = ValidatePath(dir);
+                if (string.IsNullOrEmpty(filePath))
+                {
+                    continue;
+                }
+                else
+                {
+                    dotnetPath = filePath;
+                    break;
+                }
```
ladipro, line 239: nit:

```suggestion
                if (!string.IsNullOrEmpty(filePath))
                {
                    dotnetPath = filePath;
                    break;
                }
```

ladipro (changes requested):

src/MSBuildLocator/DotNetSdkLocationHelper.cs
ladipro: `ResolveDotnetPath` is the more interesting one. I have left a comment to move this off of the static constructor also.

src/MSBuildLocator/DotNetSdkLocationHelper.cs
```diff
@@ -79,134 +69,206 @@ public static IEnumerable<VisualStudioInstance> GetInstances(string workingDirec
...
             {
-                // We will generally find the dotnet exe on the path, but on linux, it is often just a 'dotnet' symlink (possibly even to more symlinks) that we have to resolve
-                // to the real dotnet executable.
-                // This will work as often as just invoking dotnet from the command line, but we can be more confident in finding a dotnet executable by following
-                // https://github.com/dotnet/designs/blob/main/accepted/2021/install-location-per-architecture.md
-                // This can be done using the nethost library. We didn't do this previously, so I did not implement this extension.
-                foreach (string dir in Environment.GetEnvironmentVariable("PATH").Split(Path.PathSeparator))
+                // Agreed to load hostfxr from the highest version
+                var hostFxrAssemblyDirectory = Directory.GetDirectories(hostFxrRoot)
+                    .OrderByDescending(d => d)
```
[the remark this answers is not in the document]
ladipro: Ah, still not good unfortunately. The `Max` method finds the maximum in lexicographical order. We need something that understand semantic versioning, e.g. "10.0.100" is more than "7.0.200", which is more than "7.0.100", ..

src/MSBuildLocator/DotNetSdkLocationHelper.cs
```diff
@@ -75,138 +73,227 @@ public static IEnumerable<VisualStudioInstance> GetInstances(string workingDirec
...
+            if (Directory.Exists(hostFxrRoot))
             {
-                // We will generally find the dotnet exe on the path, but on linux, it is often just a 'dotnet' symlink (possibly even to more symlinks) that we have to resolve
-                // to the real dotnet executable.
-                // This will work as often as just invoking dotnet from the command line, but we can be more confident in finding a dotnet executable by following
-                // https://github.com/dotnet/designs/blob/main/accepted/2021/install-location-per-architecture.md
-                // This can be done using the nethost library. We didn't do this previously, so I did not implement this extension.
-                foreach (string dir in Environment.GetEnvironmentVariable("PATH").Split(Path.PathSeparator))
+                // Agreed to load hostfxr from the highest version
+                var hostFxrAssemblyDirectory = Directory.GetDirectories(hostFxrRoot).Max();
```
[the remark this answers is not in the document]
YuliiaKovalova: I have found a proper lib for that: NuGet.Versioning

It successfully parses and sorts folders with names like this
"8.0.0-preview.3.23174.8"

src/MSBuildLocator/DotNetSdkLocationHelper.cs
```diff
@@ -75,138 +73,227 @@ public static IEnumerable<VisualStudioInstance> GetInstances(string workingDirec
...
+            {
+                if (dotnetPaths[i] != bestSDK)
+                {
+                    yield return dotnetPaths[i];
+                }
+            }
         }
 
-        private static string FindDotnetFromEnvironmentVariable(string environmentVariable, string exeName)
+        private static void LoadHostFxr()
```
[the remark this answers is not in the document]
YuliiaKovalova: Added

src/MSBuildLocator/DotNetSdkLocationHelper.cs
```diff
@@ -75,138 +75,229 @@ public static IEnumerable<VisualStudioInstance> GetInstances(string workingDirec
...
-                        return File.Exists(fullPathToDotnetFromRoot) ? Path.GetDirectoryName(fullPathToDotnetFromRoot) : null;
-                    }
-
-                    return dotnet_root;
+                    yield return dotnetPaths[i];
                 }
             }
 
-            return null;
+            RemoveUnmanagedDllResolver();
```
ladipro, line 107: nit: Wrap in `finally` to make sure it runs even if `GetAllAvailableSDKs` throws.

ladipro (approved):
Looks good! One minor comment inline and please obtain at least one more approval.

src/MSBuildLocator/DotNetSdkLocationHelper.cs
```diff
@@ -75,138 +75,234 @@ public static IEnumerable<VisualStudioInstance> GetInstances(string workingDirec
...
 
-            string[] paths = null;
-            rc = NativeMethods.hostfxr_get_available_sdks(exe_dir: dotnetPath, result: (key, value) =>
+            return dotnetPath;
+        }
+
+        private static string? GetDotnetPathFromROOT()
+        {
+            // 32-bit architecture has (x86) suffix
+            string envVarName = (IntPtr.Size == 4) ? "DOTNET_ROOT(x86)" : "DOTNET_ROOT";
```
rainersigwald, line 223: Do we need to worry about ARM32?

rainersigwald (changes requested):
The only thing blocking IMO is the NuGet dependency at the package level.

src/MSBuildLocator/DotNetSdkLocationHelper.cs
```diff
@@ -75,138 +75,234 @@ public static IEnumerable<VisualStudioInstance> GetInstances(string workingDirec
...
             }
-            else if (IntPtr.Size == 8)
+        }
+
+        private static IntPtr HostFxrResolver(Assembly assembly, string libraryName)
+        {
+            var hostFxrLibName = "libhostfxr";
+            var libExtention = RuntimeInformation.IsOSPlatform(OSPlatform.OSX) ? "dylib" : "so";
+
+            if (!hostFxrLibName.Equals(libraryName, StringComparison.OrdinalIgnoreCase))
```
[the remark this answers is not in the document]
YuliiaKovalova: removed StringComparison.OrdinalIgnoreCase

src/MSBuildLocator/Utils/SemanticVersionParser.cs
```diff
@@ -0,0 +1,162 @@
+﻿#if NETCOREAPP
+
+using System;
+using System.Linq;
+
+namespace Microsoft.Build.Locator
+{
+    internal class SemanticVersionParser
```
rainersigwald, line 8: Please attribute this to the version of the NuGet code you looked at (in case we need to follow some changes they make)

rainersigwald, line 8: nit: `static`?

src/MSBuildLocator/DotNetSdkLocationHelper.cs
```diff
@@ -133,23 +133,24 @@ private static IntPtr HostFxrResolver(Assembly assembly, string libraryName)
...
             var hostFxrRoot = Path.Combine(DotnetPath.Value, "host", "fxr");
             if (Directory.Exists(hostFxrRoot))
             {
-                // Agreed to load hostfxr from the highest version
+                var versionParser = new SemanticVersionParser();
+                // Load hostfxr from the highest version, because it should be backward-compatible
                 var hostFxrAssemblyDirectory = Directory.GetDirectories(hostFxrRoot)
-                            .Select(str => NuGetVersion.Parse(str))
+                            .Select(str => versionParser.TryParse(str, out var version) ? version : null)
                             .Max();
```
rainersigwald, line 148: Nit: if you used `.OrderBy().First()` you would have a `DirectoryInfo` you could call `GetFiles` on which should be slightly more efficient.

rainersigwald (approved):

src/MSBuildLocator/DotNetSdkLocationHelper.cs
```diff
@@ -75,138 +74,236 @@ public static IEnumerable<VisualStudioInstance> GetInstances(string workingDirec
...
+
+        private static void ModifyUnmanagedDllResolver(Action<AssemblyLoadContext> resolverAction)
+        {
+            // For Windows hostfxr is loaded in the process.
+            if (!IsWindows)
             {
-                // 32-bit architecture
-                dotnetPath ??= FindDotnetFromEnvironmentVariable("DOTNET_ROOT(x86)", exeName);
+                var loadContext = AssemblyLoadContext.GetLoadContext(Assembly.GetExecutingAssembly());
+                if (loadContext != null)
```
Forgind, line 124: When is loadContext null? My default assumption would be that there's always at least the default context...is that wrong?

src/MSBuildLocator/DotNetSdkLocationHelper.cs
```diff
@@ -75,138 +74,236 @@ public static IEnumerable<VisualStudioInstance> GetInstances(string workingDirec
...
+                {
+                    resolverAction(loadContext);
+                }
             }
-            else if (IntPtr.Size == 8)
+        }
+
+        private static IntPtr HostFxrResolver(Assembly assembly, string libraryName)
+        {
+            var hostFxrLibName = "libhostfxr";
```
Forgind, line 133: I thought you'd implied it could be libhostfxr or hostfxr by having a regex accepting either—was that wrong?

src/MSBuildLocator/DotNetSdkLocationHelper.cs
```diff
@@ -75,138 +74,236 @@ public static IEnumerable<VisualStudioInstance> GetInstances(string workingDirec
...
-                // to the real dotnet executable.
-                // This will work as often as just invoking dotnet from the command line, but we can be more confident in finding a dotnet executable by following
-                // https://github.com/dotnet/designs/blob/main/accepted/2021/install-location-per-architecture.md
-                // This can be done using the nethost library. We didn't do this previously, so I did not implement this extension.
-                foreach (string dir in Environment.GetEnvironmentVariable("PATH").Split(Path.PathSeparator))
+                var versionParser = new SemanticVersionParser();
+                // Load hostfxr from the highest version, because it should be backward-compatible
+                var hostFxrAssemblyDirectory = Directory.GetDirectories(hostFxrRoot)
+                            .Select(str => versionParser.TryParse(str, out var version) ? version : null)
+                            .Max();
```
Forgind, line 148: I think you can combine these with:
```suggestion
                            .Max(str => versionParser.TryParse(str, out var version) ? version : null);
```

Though you might need to replace null with max version?

src/MSBuildLocator/DotNetSdkLocationHelper.cs
```diff
@@ -75,138 +74,236 @@ public static IEnumerable<VisualStudioInstance> GetInstances(string workingDirec
...
+
+                if (hostFxrAssemblyDirectory != null && !string.IsNullOrEmpty(hostFxrAssemblyDirectory.OriginalValue))
                 {
-                    string filePath = Path.Combine(dir, exeName);
-                    if (File.Exists(filePath))
+                    var hostfxrAssembly = Directory.GetFiles(hostFxrAssemblyDirectory.OriginalValue)
+                        .Where(filePath => filePath.Equals(Path.Combine(hostFxrLibName, libExtention)))
+                        .FirstOrDefault();
```
Forgind, line 154: Similarly:
```suggestion
                        .FirstOrDefault(filePath => filePath.Equals(Path.Combine(hostFxrLibName, libExtention)));
```

src/MSBuildLocator/DotNetSdkLocationHelper.cs
```diff
@@ -75,138 +74,236 @@ public static IEnumerable<VisualStudioInstance> GetInstances(string workingDirec
...
+        private static string ResolveDotnetPath()
+        {
+            string? dotnetPath = GetDotnetPathFromROOT();
+
+            if (string.IsNullOrEmpty(dotnetPath))
             {
-                throw new InvalidOperationException("Failed to find an appropriate version of .NET Core MSBuild. Call to hostfxr_resolve_sdk2 failed. There may be more details in stderr.");
+                string? dotnetExePath = GetCurrentProcessPath();
+                var isRunFromDotnetExecutable = !string.IsNullOrEmpty(dotnetExePath) 
+                    && Path.GetFileName(dotnetExePath).Equals(ExeName, StringComparison.InvariantCultureIgnoreCase);
```
Forgind, line 199: This looks like something that should be platform-specific, right?

src/MSBuildLocator/Utils/VersionComparer.cs
```diff
@@ -0,0 +1,151 @@
+﻿using System;
```
Forgind, line 1: These files need copyright headers.

src/MSBuildLocator/DotNetSdkLocationHelper.cs
```diff
@@ -75,138 +74,236 @@ public static IEnumerable<VisualStudioInstance> GetInstances(string workingDirec
...
+                
+                if (isRunFromDotnetExecutable)
+                {
+                    dotnetPath = Path.GetDirectoryName(dotnetExePath);
+                }
+                else
+                {
+                    dotnetPath = FindDotnetPathFromEnvVariable("DOTNET_HOST_PATH") 
+                        ?? FindDotnetPathFromEnvVariable("DOTNET_MSBUILD_SDK_RESOLVER_CLI_DIR")
+                        ?? GetDotnetPathFromPATH();
```
Forgind, line 209: We should have one last place where we just look in the expected places, right? Like /usr/bin/dotnet or whatever on linux?

src/MSBuildLocator/DotNetSdkLocationHelper.cs
YuliiaKovalova: Probably runtime has some cases  for returning null, since SDK also validates it 
https://github.com/dotnet/sdk/blob/285888bebe0735d3ad8a4cd0dc1cd1037ba17ed6/src/ApiCompat/Microsoft.DotNet.ApiCompat.Shared/RoslynResolver.cs#L26

src/MSBuildLocator/DotNetSdkLocationHelper.cs
YuliiaKovalova: I have checked that for osx and linux it always has this name.

src/MSBuildLocator/DotNetSdkLocationHelper.cs
YuliiaKovalova: It's an execution context-specific - if MsBuild.Locator is invoked in the scope of dotnet.exe, isRunFromDotnetExecutable is true.

src/MSBuildLocator/DotNetSdkLocationHelper.cs
YuliiaKovalova: Chet listed the lookup sequence - based on the execution context of MsBuild.Locator, we check different places. More details are in the PR description.

src/MSBuildLocator/DotNetSdkLocationHelper.cs
```diff
@@ -75,138 +74,233 @@ public static IEnumerable<VisualStudioInstance> GetInstances(string workingDirec
...
+                            .Max(str => SemanticVersionParser.TryParse(str, out var version) ? version : null);
+
+                if (hostFxrAssemblyDirectory != null && !string.IsNullOrEmpty(hostFxrAssemblyDirectory.OriginalValue))
                 {
-                    string filePath = Path.Combine(dir, exeName);
-                    if (File.Exists(filePath))
+                    var hostfxrAssembly = Directory.GetFiles(hostFxrAssemblyDirectory.OriginalValue)
+                        .FirstOrDefault(filePath => filePath.Equals(Path.Combine(hostFxrLibName, libExtention)));
+
+                    if (hostfxrAssembly != null)
```
ladipro, line 153: This looks wrong.
- `Path.Combine("name", "ext")` returns `name/ext`.
- It is wasteful to create a new string on each lambda invocation.
- Now that the comparison is case sensitive, there doesn't seem to be a reason to enumerate - we know what the expected name is.

```suggestion
                    var hostfxrAssembly = Path.Combine(hostFxrAssemblyDirectory.OriginalValue, Path.ChangeExtension(hostFxrLibName, libExtension));
                    if (File.Exists(hostFxrAssembly))
```

src/MSBuildLocator/DotNetSdkLocationHelper.cs
Forgind: I meant whether it should be IgnoreCase or not should be platform-specific?

PR #231, opened 2023-08-23
upgrade core version to net6.0 

YuliiaKovalova:
Upgrading to net6.0 uncovered that tests previously were run against framework46 and a bug introduced by respecting DOTNET_ROOT was missed.

This PR addresses both of these issues.

### DIFF
--- a/src/MSBuildLocator/DotNetSdkLocationHelper.cs
+++ b/src/MSBuildLocator/DotNetSdkLocationHelper.cs
@@ -17,7 +17,7 @@ namespace Microsoft.Build.Locator
         private static readonly Regex DotNetBasePathRegex = new Regex("Base Path:(.*)$", RegexOptions.Multiline);
         private static readonly Regex VersionRegex = new Regex(@"^(\d+)\.(\d+)\.(\d+)", RegexOptions.Multiline);
         private static readonly Regex SdkRegex = new Regex(@"(\S+) \[(.*?)]$", RegexOptions.Multiline);
-        private static bool IsWindows = RuntimeInformation.IsOSPlatform(OSPlatform.Windows);
+        private static readonly bool IsWindows = RuntimeInformation.IsOSPlatform(OSPlatform.Windows);
 
         public static VisualStudioInstance GetInstance(string dotNetSdkPath)
         {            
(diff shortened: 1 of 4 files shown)


src/MSBuildLocator/DotNetSdkLocationHelper.cs
```diff
@@ -17,6 +17,7 @@ internal static class DotNetSdkLocationHelper
         private static readonly Regex DotNetBasePathRegex = new Regex("Base Path:(.*)$", RegexOptions.Multiline);
         private static readonly Regex VersionRegex = new Regex(@"^(\d+)\.(\d+)\.(\d+)", RegexOptions.Multiline);
         private static readonly Regex SdkRegex = new Regex(@"(\S+) \[(.*?)]$", RegexOptions.Multiline);
+        private static bool IsWindows = RuntimeInformation.IsOSPlatform(OSPlatform.Windows);
```
ladipro, line 20: nit: Can the field be `readonly`?

src/MSBuildLocator/DotNetSdkLocationHelper.cs
```diff
@@ -98,7 +99,13 @@ private static string FindDotnetFromEnvironmentVariable(string environmentVariab
                 string fullPathToDotnetFromRoot = Path.Combine(dotnet_root, exeName);
                 if (File.Exists(fullPathToDotnetFromRoot))
                 {
-                    return realpath(fullPathToDotnetFromRoot) ?? fullPathToDotnetFromRoot;
+                    if (!IsWindows)
+                    {
+                        fullPathToDotnetFromRoot = realpath(fullPathToDotnetFromRoot) ?? fullPathToDotnetFromRoot;
+                        return File.Exists(fullPathToDotnetFromRoot) ? dotnet_root : null;
```
ladipro, line 105: If `fullPathToDotnetFromRoot` was assigned the real path on the previous line, wouldn't we want to return the directory part of the new real path, rather than `dotnet_root` which contains the symlink?

YuliiaKovalova: I agree. Fixed it in one more place: https://github.com/microsoft/MSBuildLocator/blob/20af00a8a63c4ffd366aeed035c7c1c960f183a8/src/MSBuildLocator/DotNetSdkLocationHelper.cs#L149

ladipro (approved):

PR #234, opened 2023-08-25
apply code style rules

YuliiaKovalova:
apply code style rules (the file is taken from MSBuild project).

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,400 @@
+# editorconfig.org
+
+# top-most EditorConfig file
+root = true
+
+# Default settings:
+# A newline ending every file
+# Use 4 spaces as indentation
+[*]
+insert_final_newline = true
+indent_style = space
+indent_size = 4
+trim_trailing_whitespace = true
+
+[project.json]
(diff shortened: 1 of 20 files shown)


.editorconfig
```diff
@@ -0,0 +1,400 @@
...
+# CA1852: Seal internal types
+dotnet_diagnostic.ca1852.severity = warning
+
+# RS0037: Enable tracking of nullability of reference types in the declared API
+# Our API is not annotated but new classes get nullable enabled so disable this.
+# We'd be happy if everything was annotated and this could be removed.
+dotnet_diagnostic.RS0037.severity = none
+
+# License header
+file_header_template = Licensed to the .NET Foundation under one or more agreements.\nThe .NET Foundation licenses this file to you under the MIT license.
```
rainersigwald, line 169: I don't think this is right for this repo, which we've had as Microsoft-owned the whole time.

YuliiaKovalova: fixed

rainersigwald (approved):

PR #236, opened 2023-08-26
Fix hostfx resolving issue in some Mac machines

lifengl:
This PR is to fix hostfx resolver issue on some Mac machines. There are a few issues fixed in this PR:

 1, the libraryName is actually the string from DllImport metadata, so it would be hostfxr
 2, enumerate directory will get a list of full path, the semantic version parser doesn't seem to handle it.
 3, DOTNET_HOST_PATH is actually path to the dotnet command file, which is different from DOTNET_ROOT, that was a misunderstanding in the code which leads it to set it incorrectly.

This PR is to address:

https://github.com/dotnet/msbuild/issues/9038

### DIFF
--- a/src/MSBuildLocator/DotNetSdkLocationHelper.cs
+++ b/src/MSBuildLocator/DotNetSdkLocationHelper.cs
@@ -131,14 +131,16 @@ namespace Microsoft.Build.Locator
 
         private static IntPtr HostFxrResolver(Assembly assembly, string libraryName)
         {
-            string hostFxrLibName = "libhostfxr";
-            string libExtension = RuntimeInformation.IsOSPlatform(OSPlatform.OSX) ? "dylib" : "so";
-
             // the DllImport hardcoded the name as hostfxr.
-            if (!hostFxrLibName.Equals(libraryName, StringComparison.OrdinalIgnoreCase) && !libraryName.Equals("hostfxr", StringComparison.OrdinalIgnoreCase))
+            if (!libraryName.Equals(NativeMethods.HostFxrName, StringComparison.Ordinal))
             {
                 return IntPtr.Zero;
             }
+
+            string hostFxrLibName =
+                RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ?
+                "hostfxr.dll" :
+                RuntimeInformation.IsOSPlatform(OSPlatform.OSX) ? "libhostfxr.dylib" : "libhostfxr.so";
 
             string hostFxrRoot = Path.Combine(DotnetPath.Value, "host", "fxr");
             if (Directory.Exists(hostFxrRoot))
@@ -151,48 +153,18 @@ namespace Microsoft.Build.Locator
                 };
 
                 // Load hostfxr from the highest version, because it should be backward-compatible
-                SemanticVersion? hostFxrVersion;
-                
-                try
+                if (fileEnumerable.Max() is SemanticVersion hostFxrVersion)
                 {
-                    hostFxrVersion = fileEnumerable.Max();
-                }
-                catch (ArgumentException)
-                {
-                    // LINQ Max throws when the list is empty.
-                    Console.Error.WriteLine($"No child folder was found in '{hostFxrRoot}'.");
-                    return IntPtr.Zero;
-                }
-
-                if (hostFxrVersion is not null)
-                {
-                    string hostFxrAssembly = Path.Combine(hostFxrRoot, hostFxrVersion.OriginalValue, Path.ChangeExtension(hostFxrLibName, libExtension));
-
-                    if (File.Exists(hostFxrAssembly))
-                    {
-                        if (NativeLibrary.TryLoad(hostFxrAssembly, out var handle))
-                        {
-                            return handle;
-                        }
-
-                        Console.Error.WriteLine($"'{hostFxrAssembly}' cannot be loaded.");
-                    }
-                    else
-                    {
-                        Console.Error.WriteLine($"hostfxr file '{hostFxrAssembly}' cannot be found.");
-                    }
-                }
-                else
-                {
-                    Console.Error.WriteLine($"No runtime version was found in '{hostFxrRoot}'.");
+                    string hostFxrAssembly = Path.Combine(hostFxrRoot, hostFxrVersion.OriginalValue, hostFxrLibName);
+                    return NativeLibrary.Load(hostFxrAssembly);
                 }
             }
-            else
-            {
-                Console.Error.WriteLine($"HostFxr folder is missing: '{hostFxrRoot}'");
-            }
 
-            return IntPtr.Zero;
+            string error = $".NET SDK cannot be resolved, because {hostFxrLibName} cannot be found inside {hostFxrRoot}." +
+                Environment.NewLine +
+                $"This might indicate a corrupted SDK installation on the machine.";
+
+            throw new InvalidOperationException(error);
         }
 
         private static string SdkResolutionExceptionMessage(string methodName) => $"Failed to find all versions of .NET Core MSBuild. Call to {methodName}. There may be more details in stderr.";
--- a/src/MSBuildLocator/DotNetSdkLocationHelper.cs
+++ b/src/MSBuildLocator/DotNetSdkLocationHelper.cs
@@ -170,7 +170,12 @@ namespace Microsoft.Build.Locator
 
                     if (File.Exists(hostFxrAssembly))
                     {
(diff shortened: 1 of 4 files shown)


src/MSBuildLocator/DotNetSdkLocationHelper.cs
```diff
@@ -130,24 +131,48 @@ private static void ModifyUnmanagedDllResolver(Action<AssemblyLoadContext> resol
...
             var libExtension = RuntimeInformation.IsOSPlatform(OSPlatform.OSX) ? "dylib" : "so";
 
-            if (!hostFxrLibName.Equals(libraryName))
+            // the DllImport hardcoded the name as hostfxr.
+            if (!hostFxrLibName.Equals(libraryName, StringComparison.Ordinal) && !libraryName.Equals("hostfxr", StringComparison.Ordinal))
+            {
+                return IntPtr.Zero;
+            }
+
+            string? dotnetFolderPath = Path.GetDirectoryName(DotnetPath.Value);
```
YuliiaKovalova, line 145: this check is present in https://github.com/lifengl/MSBuildLocator/blob/937a06d622caf3a0fd0326479375700ea1e93932/src/MSBuildLocator/DotNetSdkLocationHelper.cs#L234C18-L234C18

lifengl: actually, I misunderstood the issue, and DotNetPath is the folder path, so I should revert that change. The problem here is DOTNET_HOST_PATH should point to the file, not folder, which lead to some confusions. After fixing that, it seems to work. I also removed excessive logging code added earlier.

lifengl:
cc @chzuluag

lifengl:
+ @ladipro

src/MSBuildLocator/DotNetSdkLocationHelper.cs
```diff
@@ -130,30 +131,65 @@ private static void ModifyUnmanagedDllResolver(Action<AssemblyLoadContext> resol
...
         private static IntPtr HostFxrResolver(Assembly assembly, string libraryName)
         {
-            var hostFxrLibName = "libhostfxr";
-            var libExtension = RuntimeInformation.IsOSPlatform(OSPlatform.OSX) ? "dylib" : "so";
+            string hostFxrLibName = "libhostfxr";
+            string libExtension = RuntimeInformation.IsOSPlatform(OSPlatform.OSX) ? "dylib" : "so";
 
-            if (!hostFxrLibName.Equals(libraryName))
+            // the DllImport hardcoded the name as hostfxr.
+            if (!hostFxrLibName.Equals(libraryName, StringComparison.OrdinalIgnoreCase) && !libraryName.Equals("hostfxr", StringComparison.OrdinalIgnoreCase))
```
ladipro, line 138: nit: If `libraryName` we get here comes from the `DllImport` attribute verbatim, why not check for it with a straight equals? I.e. no need for case insensitivity or detecting the other name.

```suggestion
            // the DllImport hardcoded the name as hostfxr.
            if (!libraryName.Equals("hostfxr"))
```

src/MSBuildLocator/DotNetSdkLocationHelper.cs
```diff
@@ -130,30 +131,65 @@ private static void ModifyUnmanagedDllResolver(Action<AssemblyLoadContext> resol
...
-            var hostFxrRoot = Path.Combine(DotnetPath.Value, "host", "fxr");
+            string hostFxrRoot = Path.Combine(DotnetPath.Value, "host", "fxr");
             if (Directory.Exists(hostFxrRoot))
             {
+                var fileEnumerable = new FileSystemEnumerable<SemanticVersion?>(
+                    directory: hostFxrRoot,
+                    transform: static (ref FileSystemEntry entry) => SemanticVersionParser.TryParse(entry.FileName.ToString(), out var version) ? version : null)
+                {
+                    ShouldIncludePredicate = static (ref FileSystemEntry entry) => entry.IsDirectory
+                };
```
ladipro, line 151: ❤️

src/MSBuildLocator/DotNetSdkLocationHelper.cs
```diff
@@ -211,7 +263,7 @@ private static string ResolveDotnetPath()
                 throw new InvalidOperationException("Could not find the dotnet executable. Is it set on the DOTNET_ROOT?");
             }
 
-            SetEnvironmentVariableIfEmpty("DOTNET_HOST_PATH", dotnetPath);
+            SetEnvironmentVariableIfEmpty("DOTNET_HOST_PATH", Path.Combine(dotnetPath, ExeName));
```
YuliiaKovalova, line 266: we need to set dotnet path only in DOTNET_HOST_PATH

src/MSBuildLocator/DotNetSdkLocationHelper.cs
```diff
@@ -130,30 +131,65 @@ private static void ModifyUnmanagedDllResolver(Action<AssemblyLoadContext> resol
...
+                try
+                {
+                    hostFxrVersion = fileEnumerable.Max();
+                }
+                catch (ArgumentException)
+                {
+                    // LINQ Max throws when the list is empty.
+                    Console.Error.WriteLine($"No child folder was found in '{hostFxrRoot}'.");
+                    return IntPtr.Zero;
+                }
```
ladipro, line 165: `Enumerable.Max` is documented to throw `InvalidOperationException` if the sequence is empty. Does not seem to apply to the overload being called here, though. `Max` just returns null: https://dotnetfiddle.net/JRnHee

```suggestion
                SemanticVersion? hostFxrVersion = fileEnumerable.Max();
```

YuliiaKovalova (approved):

src/MSBuildLocator/DotNetSdkLocationHelper.cs
```diff
@@ -130,30 +131,65 @@ private static void ModifyUnmanagedDllResolver(Action<AssemblyLoadContext> resol
...
 
                     if (File.Exists(hostFxrAssembly))
                     {
-                        return NativeLibrary.TryLoad(hostFxrAssembly, out var handle) ? handle : IntPtr.Zero;
+                        if (NativeLibrary.TryLoad(hostFxrAssembly, out var handle))
+                        {
+                            return handle;
+                        }
+
+                        Console.Error.WriteLine($"'{hostFxrAssembly}' cannot be loaded.");
```
ladipro, line 178: I think it's much better if library code throws exceptions rather than prints to stderr. As far as I can tell exceptions thrown here should bubble up to the component that calls MSBuildLocator. Would that work for your scenario?

lifengl:
Both Christian and CTI has verified the PR fixed the issue on Mac (which can be produced on their x64 mac machine earlier.)

ladipro (commented):
Thank you, I've left a few comments inline. I understand we're under time pressure but please consider throwing exceptions rather than printing diagnostics to stderr.

src/MSBuildLocator/DotNetSdkLocationHelper.cs
lifengl: my bad, i misread the document. There are indeed different behavior for different variations of this method...

src/MSBuildLocator/DotNetSdkLocationHelper.cs
lifengl: fair point. I just don't want to change the original behavior, but since it is a recent change, I guess no other logic would depend on the original path.  I will do that, if it works.

src/MSBuildLocator/DotNetSdkLocationHelper.cs
lifengl: hmm... somehow I got an impression that this library actually logs errors in this way, but must be a mistake. It is likely from another place related to SDK resolution.  I didn't try to throw, because I wonder this can increase risks as this method is chained to global unmanaged dll loading event.  It matches the logic to use TryLoad, instead of Load. But on the other hand, it can throw anyway from DotnetPath.Value, so I guess adopting throwing error is reasonable.

it would be nice if this code can have some Trace logging logic. But no reason to expand it further at this point.

src/MSBuildLocator/DotNetSdkLocationHelper.cs
lifengl: yeah, Christian verified it, so I deleted that piece.

ladipro (approved):

lifengl:
@YuliiaKovalova , @ladipro : can your team help to merge it when you feel it is ready.  (I don't have permission to do it.)

PR #243, opened 2023-09-14
add probing of all available path candidates + return missed DOTNET_HOST_PATH set change

YuliiaKovalova:
add probing of all available path candidates + return missed DOTNET_HOST_PATH set change from https://github.com/microsoft/MSBuildLocator/pull/235/files

### DIFF
--- a/src/MSBuildLocator/DotNetSdkLocationHelper.cs
+++ b/src/MSBuildLocator/DotNetSdkLocationHelper.cs
@@ -22,7 +22,7 @@ namespace Microsoft.Build.Locator
         private static readonly Regex s_versionRegex = new(@"^(\d+)\.(\d+)\.(\d+)", RegexOptions.Multiline);
         private static readonly bool s_isWindows = RuntimeInformation.IsOSPlatform(OSPlatform.Windows);
         private static readonly string s_exeName = s_isWindows ? "dotnet.exe" : "dotnet";
-        private static readonly Lazy<string> s_dotnetPath = new(() => ResolveDotnetPath());
+        private static readonly Lazy<IList<string>> s_dotnetPathCandidates = new(() => ResolveDotnetPathCandidates());
 
         public static VisualStudioInstance? GetInstance(string dotNetSdkPath)
         {
@@ -139,24 +139,27 @@ namespace Microsoft.Build.Locator
                 return IntPtr.Zero;
             }
 
-            string hostFxrRoot = Path.Combine(s_dotnetPath.Value, "host", "fxr");
-            if (Directory.Exists(hostFxrRoot))
+            foreach (string dotnetPath in s_dotnetPathCandidates.Value)
             {
-                // Load hostfxr from the highest version, because it should be backward-compatible
-                SemanticVersion? hostFxrAssemblyDirectory = Directory.GetDirectories(hostFxrRoot)
-                            .Max(str => SemanticVersionParser.TryParse(str, out SemanticVersion? version) ? version : null);
-
-                if (hostFxrAssemblyDirectory != null && !string.IsNullOrEmpty(hostFxrAssemblyDirectory.OriginalValue))
+                string hostFxrRoot = Path.Combine(dotnetPath, "host", "fxr");
+                if (Directory.Exists(hostFxrRoot))
                 {
-                    string hostFxrAssembly = Path.Combine(hostFxrAssemblyDirectory.OriginalValue, Path.ChangeExtension(hostFxrLibName, libExtension));
+                    // Load hostfxr from the highest version, because it should be backward-compatible
+                    SemanticVersion? hostFxrAssemblyDirectory = Directory.GetDirectories(hostFxrRoot)
+                                .Max(str => SemanticVersionParser.TryParse(str, out SemanticVersion? version) ? version : null);
 
-                    if (File.Exists(hostFxrAssembly))
+                    if (hostFxrAssemblyDirectory != null && !string.IsNullOrEmpty(hostFxrAssemblyDirectory.OriginalValue))
                     {
-                        return NativeLibrary.TryLoad(hostFxrAssembly, out IntPtr handle) ? handle : IntPtr.Zero;
+                        string hostFxrAssembly = Path.Combine(hostFxrAssemblyDirectory.OriginalValue, Path.ChangeExtension(hostFxrLibName, libExtension));
+
+                        if (File.Exists(hostFxrAssembly))
+                        {
+                            return NativeLibrary.TryLoad(hostFxrAssembly, out IntPtr handle) ? handle : IntPtr.Zero;
+                        }
                     }
                 }
             }
-
+           
             return IntPtr.Zero;
         }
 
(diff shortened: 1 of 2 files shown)


src/MSBuildLocator/DotNetSdkLocationHelper.cs
```diff
@@ -169,39 +172,55 @@ private static IntPtr HostFxrResolver(Assembly assembly, string libraryName)
...
             }
 
-            return string.IsNullOrEmpty(dotnetPath)
+            pathCandidates.Add(FindDotnetPathFromEnvVariable("DOTNET_HOST_PATH"));
+            pathCandidates.Add(FindDotnetPathFromEnvVariable("DOTNET_MSBUILD_SDK_RESOLVER_CLI_DIR"));
+            pathCandidates.Add(GetDotnetPathFromPATH());
+
+            IList<string> filteredPathCandidates = pathCandidates.Where(pk => !string.IsNullOrEmpty(pk))
+                .Select(s => s!)
+                .ToList();
```
ladipro, line 219: nit: You can do this filtering right when adding to the list for a bit of extra efficiency. Maybe with a local function like:

```C#
    // ...
    
    AddIfValid(GetDotnetPathFromROOT());
    // ...
    AddIfValid(FindDotnetPathFromEnvVariable("DOTNET_HOST_PATH"));
    AddIfValid(FindDotnetPathFromEnvVariable("DOTNET_MSBUILD_SDK_RESOLVER_CLI_DIR"));
    AddIfValid(GetDotnetPathFromPATH());

    // ...

    void AddIfValid(string? path)
    {
        if (!string.IsNullOrEmpty(path))
        {
            pathCandidates.Add(path);
        }
    }
```

src/MSBuildLocator/DotNetSdkLocationHelper.cs
```diff
@@ -169,39 +172,55 @@ private static IntPtr HostFxrResolver(Assembly assembly, string libraryName)
...
+
+            return filteredPathCandidates.Count == 0
                 ? throw new InvalidOperationException("Could not find the dotnet executable. Is it set on the DOTNET_ROOT?")
```
ladipro, line 222: Should the exception message be updated?

src/MSBuildLocator/DotNetSdkLocationHelper.cs
```diff
@@ -244,10 +263,18 @@ private static string ResolveDotnetPath()
...
+                {
+                    break;
+                }
+            }
 
             // Errors are automatically printed to stderr. We should not continue to try to output anything if we failed.
-            return rc != 0
+            return resolvedPaths == null
                 ? throw new InvalidOperationException(SdkResolutionExceptionMessage(nameof(NativeMethods.hostfxr_get_available_sdks)))
                 : resolvedPaths ?? Array.Empty<string>();
```
ladipro, line 279: nit: This `?? Array.Empty<string>()` piece does not do anything now.

src/MSBuildLocator/DotNetSdkLocationHelper.cs
```diff
@@ -139,24 +139,27 @@ private static IntPtr HostFxrResolver(Assembly assembly, string libraryName)
...
 
-                    if (File.Exists(hostFxrAssembly))
+                    if (hostFxrAssemblyDirectory != null && !string.IsNullOrEmpty(hostFxrAssemblyDirectory.OriginalValue))
                     {
-                        return NativeLibrary.TryLoad(hostFxrAssembly, out IntPtr handle) ? handle : IntPtr.Zero;
+                        string hostFxrAssembly = Path.Combine(hostFxrAssemblyDirectory.OriginalValue, Path.ChangeExtension(hostFxrLibName, libExtension));
+
+                        if (File.Exists(hostFxrAssembly))
+                        {
+                            return NativeLibrary.TryLoad(hostFxrAssembly, out IntPtr handle) ? handle : IntPtr.Zero;
```
ladipro, line 157: Given the issues with running ad-hoc signed code on Mac, I wonder if we shouldn't make this part also more defensive. If `TryLoad` fails for the highest version, we could try to call one from the second highest, and so on.

Basically (pseudo-code):

```
for all dotnet location candidates:
  for all host\fxr subdirectories ordered from semantically highest version:
    if TryLoad() succeeds, return its result, otherwise continue
```

cc @lifengl

src/MSBuildLocator/DotNetSdkLocationHelper.cs
YuliiaKovalova: done

src/MSBuildLocator/DotNetSdkLocationHelper.cs
YuliiaKovalova: please check the updated message

src/MSBuildLocator/DotNetSdkLocationHelper.cs
YuliiaKovalova: done

src/MSBuildLocator/DotNetSdkLocationHelper.cs
```diff
@@ -131,35 +131,51 @@ private static void ModifyUnmanagedDllResolver(Action<AssemblyLoadContext> resol
...
+                // Check if the host/fxr directory exists
+                if (Directory.Exists(hostFxrRoot))
                 {
-                    string hostFxrAssembly = Path.Combine(hostFxrAssemblyDirectory.OriginalValue, Path.ChangeExtension(hostFxrLibName, libExtension));
-
-                    if (File.Exists(hostFxrAssembly))
+                    // Get a list of hostfxr assembly directories (e.g., 6.0.3, 7.0.1-preview.2.4)
+                    IList<SemanticVersion> hostFxrAssemblyDirs = Directory.GetDirectories(hostFxrRoot)
+                        .Select(path => SemanticVersionParser.TryParse(Path.GetFileName(path), out SemanticVersion? version) ? version : null)
+                        .Select(v => v!)
```
ladipro, line 156: The previous `Select` may return null. Shouldn't this be a `Where` filter?

src/MSBuildLocator/DotNetSdkLocationHelper.cs
```diff
@@ -169,39 +185,57 @@ private static IntPtr HostFxrResolver(Assembly assembly, string libraryName)
...
 
-            return rc != 0
+                if (rc == 0 && !string.IsNullOrEmpty(resolvedSdk))
+                {
+                    SetEnvironmentVariableIfEmpty("DOTNET_HOST_PATH", dotnetPath);
+                    break;
+                }
+            }
+            
+            return string.IsNullOrEmpty(resolvedSdk)
```
ladipro, line 205: Is the difference between this condition and the `(rc == 0 && !string.IsNullOrEmpty(resolvedSdk))` a few lines above intentional? Do we want to handle the case where `hostfxr_resolve_sdk2` returns non-zero but the callbacks sets a non-empty `resolvedSdk`? Maybe the code inside the foreach loop could return instead of breaking.

src/MSBuildLocator/DotNetSdkLocationHelper.cs
ladipro: nit: Typo "it is added **it**"

Also, should the message list all environment variable that we check here?

YuliiaKovalova: @baronfel we need your advice here :)

src/MSBuildLocator/DotNetSdkLocationHelper.cs
YuliiaKovalova: It's strange, but` .Where(v => v != null) `instead of ` .Select(v => v!)` generates [CS8619](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-messages/nullable-warnings?f1url=%3FappId%3Droslyn%26k%3Dk(CS8619)#mismatch-in-nullability-declaration:~:text=match%20implemented%20member.-,CS8619,-%2D%20Nullability%20of)

src/MSBuildLocator/DotNetSdkLocationHelper.cs
YuliiaKovalova: done

src/MSBuildLocator/DotNetSdkLocationHelper.cs
ladipro: It is tracked as a Roslyn issue: https://github.com/dotnet/roslyn/issues/37468

src/MSBuildLocator/DotNetSdkLocationHelper.cs
```diff
@@ -169,39 +186,57 @@ private static IntPtr HostFxrResolver(Assembly assembly, string libraryName)
...
 
-            return string.IsNullOrEmpty(dotnetPath)
-                ? throw new InvalidOperationException("Could not find the dotnet executable. Is it set on the DOTNET_ROOT?")
-                : dotnetPath;
+            AddIfValid(FindDotnetPathFromEnvVariable("DOTNET_HOST_PATH"));
+            AddIfValid(FindDotnetPathFromEnvVariable("DOTNET_MSBUILD_SDK_RESOLVER_CLI_DIR"));
+            AddIfValid(GetDotnetPathFromPATH());
+
+            return pathCandidates.Count == 0
+                ? throw new InvalidOperationException("Path to dotnet executable is not set. Make sure it is added either to DOTNET_HOST_PATH or PATH environment variable.")
```
baronfel, line 230: This message needs to lay out the entire set of checked variables/locations, IMO.

baronfel (approved):

src/MSBuildLocator/DotNetSdkLocationHelper.cs
YuliiaKovalova: the message is updated

PR #244, opened 2023-09-16
Revert code style changes

YuliiaKovalova:
Revert code style changes + undo all extra changes in DotNetSdkLocationHelper


ladipro (approved):

dibarbet (approved):

src/MSBuildLocator/DotNetSdkLocationHelper.cs
```diff
@@ -214,7 +214,7 @@ private static IList<string> ResolveDotnetPathCandidates()
 
             string? dotnetExePath = GetCurrentProcessPath();
             bool isRunFromDotnetExecutable = !string.IsNullOrEmpty(dotnetExePath)
-                && Path.GetFileName(dotnetExePath).Equals(s_exeName, StringComparison.InvariantCultureIgnoreCase);
+                && Path.GetFileName(dotnetExePath).Equals(ExeName, StringComparison.InvariantCultureIgnoreCase);
 
             if (isRunFromDotnetExecutable)
             {
```
lifengl, line 220: We still lost some fixes here, the issue is DOTNET_HOST_PATH points to the dotnet executable, so it cannot be done the same way as other environment variables pointing to a directory.

in the code lost in the last Style PR, the code looks like this ( I guess it needs to be integrated back as the logic is now different.  Basically, it needs to handle symbol link, and then get the directory containing the file


                    // DOTNET_HOST_PATH is pointing to the file, DOTNET_ROOT is the path of the folder
                    string? hostPath = Environment.GetEnvironmentVariable("DOTNET_HOST_PATH");
                    if (!string.IsNullOrEmpty(hostPath) && File.Exists(hostPath))
                    {
                        if (!IsWindows)
                        {
                            hostPath = realpath(hostPath) ?? hostPath;
                        }

                        dotnetPath = Path.GetDirectoryName(hostPath);
                        if (dotnetPath is not null)
                        {
                            // don't overwrite DOTNET_HOST_PATH, if we use it.
                            return dotnetPath;
                        }
                    }

lifengl (changes requested):

src/MSBuildLocator/DotNetSdkLocationHelper.cs
YuliiaKovalova: fixed

Issue #101, opened 2020-11-09
Add NuGet keyword "MSBuildLocator"


rainersigwald:
I just told @BenVillalobos to add a reference to "MSBuildLocator" and quite reasonably he typed `msbuildlocator` into NuGet search, which found [Nuke.MSBuildLocator](https://www.nuget.org/packages/Nuke.MSBuildLocator/) but not `Microsoft.Build.Locator`.

Can we put metadata into the package so it turns up under that search?

YuliiaKovalova:
@rainersigwald , could you check if the search returns now the desirable outcome:
![image](https://github.com/microsoft/MSBuildLocator/assets/95473390/8ef77122-5cbb-467c-ba1b-649d27970aee)

rainersigwald:
Boo, it looks like it's case-sensitive:
![image](https://github.com/microsoft/MSBuildLocator/assets/3347530/ea8c06fd-48d9-4009-b00d-5b531ce4a1ea)

PR #265, opened 2023-12-14
Update condition for max version limitation & add package tags

YuliiaKovalova:
## Fixes 
Updates restriction on the max version detection - previously it was capped by the environment version + add package metadata #101

### DIFF
--- a/src/MSBuildLocator/DotNetSdkLocationHelper.cs
+++ b/src/MSBuildLocator/DotNetSdkLocationHelper.cs
@@ -24,7 +24,7 @@ namespace Microsoft.Build.Locator
         private static readonly string ExeName = IsWindows ? "dotnet.exe" : "dotnet";
         private static readonly Lazy<IList<string>> s_dotnetPathCandidates = new(() => ResolveDotnetPathCandidates());
 
-        public static VisualStudioInstance? GetInstance(string dotNetSdkPath)
+        public static VisualStudioInstance? GetInstance(string dotNetSdkPath, bool allowQueryAllRuntimeVersions)
         {
             if (string.IsNullOrWhiteSpace(dotNetSdkPath) || !File.Exists(Path.Combine(dotNetSdkPath, "Microsoft.Build.dll")))
             {
@@ -52,6 +52,16 @@ namespace Microsoft.Build.Locator
                 return null;
             }
 
+            // Components of the SDK often have dependencies on the runtime they shipped with, including that several tasks that shipped
+            // in the .NET 5 SDK rely on the .NET 5.0 runtime. Assuming the runtime that shipped with a particular SDK has the same version,
+            // this ensures that we don't choose an SDK that doesn't work with the runtime of the chosen application. This is not guaranteed
+            // to always work but should work for now.
+            if (major > Environment.Version.Major || (major == Environment.Version.Major && minor > Environment.Version.Minor) 
+                && !allowQueryAllRuntimeVersions)
+            {
+                return null;
+            }
+
             return new VisualStudioInstance(
                 name: ".NET Core SDK",
                 path: dotNetSdkPath,
@@ -59,11 +69,11 @@ namespace Microsoft.Build.Locator
                 discoveryType: DiscoveryType.DotNetSdk);
         }
 
-        public static IEnumerable<VisualStudioInstance> GetInstances(string workingDirectory)
+        public static IEnumerable<VisualStudioInstance> GetInstances(string workingDirectory, bool allowQueryAllRuntimes)
         {            
             foreach (var basePath in GetDotNetBasePaths(workingDirectory))
             {
-                var dotnetSdk = GetInstance(basePath);
+                var dotnetSdk = GetInstance(basePath, allowQueryAllRuntimes);
                 if (dotnetSdk != null)
                 {
                     yield return dotnetSdk;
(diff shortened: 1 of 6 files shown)


src/MSBuildLocator/VisualStudioInstanceQueryOptions.cs
```diff
@@ -33,5 +33,11 @@ public class VisualStudioInstanceQueryOptions
         ///     Working directory to use when querying for instances. Ensure it is the project directory to pick up the right global.json.
         /// </summary>
         public string WorkingDirectory { get; set; } = Environment.CurrentDirectory;
+
+        /// <summary>
+        ///     This variable enables the removal of the existing restriction on querying installed Visual Studio instances 
+        ///     with a runtime version lower than or equal to the one in the current environment.
+        /// </summary>
+        public bool AllowQueryAllRuntimeVersions { get; set; } = Environment.GetEnvironmentVariable("DOTNET_MSBUILD_QUERY_ALL_RUNTIMES") == "1";
```
rainersigwald, line 41: What's motivating the environment variable here, instead of just requiring this setting?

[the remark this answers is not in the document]
rainersigwald: @baronfel is there a good scenario for the inverse--.NET Framework locator finding SDK installations?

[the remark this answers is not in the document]
baronfel: This would be part of the bucket of scenarios where the user is almost certainly not going to be able to _load_ the MSBuild, but they might just want to list/shell-invoke/etc the MSBuild? It could be useful, but a) wouldn't be as high of a priority in my mind, and b) would need to be behind whatever 'I promise I know this won't work if I try to in-proc this' flag/gate we've set up here.

YuliiaKovalova: I wanted to avoid extending public API. 
If adding a flag is a better option I can change.

YuliiaKovalova: fixed

src/MSBuildLocator/MSBuildLocator.cs
```diff
@@ -43,6 +43,12 @@ public static class MSBuildLocator
         /// </summary>
         public static bool IsRegistered => s_registeredHandler != null;
 
+        /// <summary>
+        ///     This flag enables the removal of the existing restriction on querying installed Visual Studio instances 
+        ///     with a runtime version lower than or equal to the one in the current environment.
+        /// </summary>
```
rainersigwald, line 49: I would like to rephrase this to be scarier :)

```suggestion
        /// <summary>
        ///     Allow discovery of .NET SDK versions that are unlikely to be successfully loaded in the current process.
        /// </summary>
        /// <remarks>
        ///     Defaults to <see langword="false"/>. Set this to <see langword="true"/> only if your application has special logic to handle loading an incompatible SDK, such as launching a new process with the target SDK's runtime.
        /// </remarks.
```

src/MSBuildLocator/DotNetSdkLocationHelper.cs
```diff
@@ -56,8 +56,8 @@ internal static class DotNetSdkLocationHelper
             // in the .NET 5 SDK rely on the .NET 5.0 runtime. Assuming the runtime that shipped with a particular SDK has the same version,
             // this ensures that we don't choose an SDK that doesn't work with the runtime of the chosen application. This is not guaranteed
             // to always work but should work for now.
-            if (major > Environment.Version.Major ||
-                (major == Environment.Version.Major && minor > Environment.Version.Minor))
+            if (major > Environment.Version.Major || (major == Environment.Version.Major && minor > Environment.Version.Minor) 
+                && !allowQueryAllRuntimeVersions)
```
rainersigwald, line 60: I can never remember boolean operator precedence rules. How would you feel about reordering this?

```suggestion
            if (!allowQueryAllRuntimeVersions && 
                (major > Environment.Version.Major || 
                 (major == Environment.Version.Major && minor > Environment.Version.Minor)))
```

rainersigwald (approved):
A couple of nits but looking good IMO. Can you give a minor-version bump in `version.json` since this is a feature addition?

src/MSBuildLocator/MSBuildLocator.cs
```diff
@@ -43,6 +43,14 @@ public static class MSBuildLocator
         /// </summary>
         public static bool IsRegistered => s_registeredHandler != null;
 
+        /// <summary>
+        ///     Allow discovery of .NET SDK versions that are unlikely to be successfully loaded in the current process.
+        /// </summary>
+        /// <remarks>
+        ///     Defaults to <see langword="false"/>. Set this to <see langword="true"/> only if your application has special logic to handle loading an incompatible SDK, such as launching a new process with the target SDK's runtime.
+        /// </remarks.
+        public static bool AllowQueryAllRuntimeVersions { get; set; } = false;
```
jasonmalinowski, line 52: Sorry I only noticed this PR now, but @YuliiaKovalova should a flag like this have been added to the VisualStudioInstanceQueryOptions rather than being a global static? Otherwise it's a bit unclear what a library like Roslyn's MSBuildWorkspace should do here since technically we're just a library and won't ever be registering MSBuild in the calling applications' process at this point -- do we set this and set it back to the old value, and pray there wasn't another thread doing something?

rainersigwald: Oh, I completely missed that move! I agree with @jasonmalinowski that this made more sense on `VisualStudioInstanceQueryOptions`.

PR #281, opened 2024-02-22
Update the place for AllowQueryAllRuntimeVersions

YuliiaKovalova:
## Fixes

No functional changes, just update the place of the flag AllowQueryAllRuntimeVersions in order to address the Roslyn's team needs. See https://github.com/microsoft/MSBuildLocator/pull/265#discussion_r1498227732 for more details.

### DIFF
--- a/src/MSBuildLocator/MSBuildLocator.cs
+++ b/src/MSBuildLocator/MSBuildLocator.cs
@@ -4,7 +4,6 @@
 using System;
 using System.Collections.Generic;
 using System.ComponentModel;
-using System.Diagnostics;
 using System.IO;
 using System.Linq;
 using System.Reflection;
@@ -12,6 +11,8 @@ using System.Text;
 
 #if NETCOREAPP
 using System.Runtime.Loader;
+#else
+using System.Diagnostics;
 #endif
 
 namespace Microsoft.Build.Locator
@@ -42,14 +43,6 @@ namespace Microsoft.Build.Locator
         ///     Gets a value indicating whether an instance of MSBuild is currently registered.
         /// </summary>
         public static bool IsRegistered => s_registeredHandler != null;
-
-        /// <summary>
-        ///     Allow discovery of .NET SDK versions that are unlikely to be successfully loaded in the current process.
-        /// </summary>
-        /// <remarks>
-        ///     Defaults to <see langword="false"/>. Set this to <see langword="true"/> only if your application has special logic to handle loading an incompatible SDK, such as launching a new process with the target SDK's runtime.
-        /// </remarks.
-        public static bool AllowQueryAllRuntimeVersions { get; set; } = false;
 
         /// <summary>
         ///     Gets a value indicating whether an instance of MSBuild can be registered.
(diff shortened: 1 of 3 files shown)


src/MSBuildLocator/MSBuildLocator.cs
```diff
@@ -43,14 +44,6 @@ public static class MSBuildLocator
         /// </summary>
         public static bool IsRegistered => s_registeredHandler != null;
 
-        /// <summary>
-        ///     Allow discovery of .NET SDK versions that are unlikely to be successfully loaded in the current process.
-        /// </summary>
-        /// <remarks>
-        ///     Defaults to <see langword="false"/>. Set this to <see langword="true"/> only if your application has special logic to handle loading an incompatible SDK, such as launching a new process with the target SDK's runtime.
-        /// </remarks.
-        public static bool AllowQueryAllRuntimeVersions { get; set; } = false;
```
jasonmalinowski, line 52: No concerns about this being a breaking change? I don't know what your API compat model is; I could imagine in your case it's just an app picking a version and that's fine.

src/MSBuildLocator/VisualStudioInstanceQueryOptions.cs
```diff
@@ -29,6 +29,16 @@ public class VisualStudioInstanceQueryOptions
...
         public DiscoveryType DiscoveryTypes { get; set; }
 
+#if NETCOREAPP
+        /// <summary>
+        ///     Allow discovery of .NET SDK versions that are unlikely to be successfully loaded in the current process.
+        /// </summary>
+        /// <remarks>
+        ///     Defaults to <see langword="false"/>. Set this to <see langword="true"/> only if your application has special logic to handle loading an incompatible SDK, such as launching a new process with the target SDK's runtime.
+        /// </remarks.
+        public bool AllowQueryAllRuntimeVersions { get; set; } = false;
```
jasonmalinowski, line 39:
```suggestion
        public bool AllowAllRuntimeVersions { get; set; } = false;
```

I admit the 'query' made it hard to read for me, but you can ignore this. (Query is implied in the type name anyways.)

jasonmalinowski (approved):
That works for us; I don't know what your breaking change policy is but we haven't consumed the other one yet so that's fine for us.

src/MSBuildLocator/MSBuildLocator.cs
rainersigwald: We've avoided breaking changes (at least against published-to-nuget.org versions) so far; we should probably keep doing so. I guess allow both?

(We should also be accumulating a list of breaking changes, I have no compunction against releasing a 2.0 with a cleaner but breaking API.)

rainersigwald (changes requested):

src/MSBuildLocator/MSBuildLocator.cs
YuliiaKovalova: Sounds good! Feel free to contribute https://github.com/microsoft/MSBuildLocator/issues/282

rainersigwald (approved):

YuliiaKovalova:
Hi @jasonmalinowski,

The package with the requested change has been just published: https://github.com/microsoft/MSBuildLocator/releases/tag/v1.7.8

PR #318, opened 2024-12-12
Update NuGet package icon.

teo-tsirpanis:
This PR updates the NuGet package icon to the same as https://github.com/dotnet/msbuild/blob/main/branding/MSBuild-NuGet-Icon.png, last updated in dotnet/msbuild#11020.


YuliiaKovalova (approved):
Thank you!

PR #329, opened 2025-03-19
Return SDKs from all dotnet location

JoeRobich:
[Review with whitespace off](https://github.com/microsoft/MSBuildLocator/pull/329/files?diff=split&w=1) as the formatter removed trailing whitespace.

The method which finds all SDKs has an early exit which stops it from finding SDKs from more than one dotnet install. This PR removes the early exit and filters the returned instances by their reported version number. This should deduplicate the SDKs.

### DIFF
--- a/src/MSBuildLocator/DotNetSdkLocationHelper.cs
+++ b/src/MSBuildLocator/DotNetSdkLocationHelper.cs
@@ -56,8 +56,8 @@ namespace Microsoft.Build.Locator
             // in the .NET 5 SDK rely on the .NET 5.0 runtime. Assuming the runtime that shipped with a particular SDK has the same version,
             // this ensures that we don't choose an SDK that doesn't work with the runtime of the chosen application. This is not guaranteed
             // to always work but should work for now.
-            if (!allowQueryAllRuntimeVersions && 
-                (major > Environment.Version.Major || 
+            if (!allowQueryAllRuntimeVersions &&
+                (major > Environment.Version.Major ||
                  (major == Environment.Version.Major && minor > Environment.Version.Minor)))
             {
                 return null;
@@ -71,13 +71,18 @@ namespace Microsoft.Build.Locator
         }
 
         public static IEnumerable<VisualStudioInstance> GetInstances(string workingDirectory, bool allowQueryAllRuntimes)
-        {            
+        {
+            HashSet<Version> versions = new();
             foreach (var basePath in GetDotNetBasePaths(workingDirectory))
             {
                 var dotnetSdk = GetInstance(basePath, allowQueryAllRuntimes);
                 if (dotnetSdk != null)
                 {
-                    yield return dotnetSdk;
+                    // Only return an SDK once, even if it's installed in multiple locations.
+                    if (versions.Add(dotnetSdk.Version))
+                    {
+                        yield return dotnetSdk;
+                    }
                 }
             }
         }
@@ -158,7 +163,7 @@ namespace Microsoft.Build.Locator
                     };
 
                     var orderedVersions = fileEnumerable.Where(v => v != null).Select(v => v!).OrderByDescending(f => f).ToList();
-                    
+
                     foreach (SemanticVersion hostFxrVersion in orderedVersions)
                     {
                         string hostFxrAssembly = Path.Combine(hostFxrRoot, hostFxrVersion.OriginalValue, hostFxrLibName);
@@ -178,7 +183,7 @@ namespace Microsoft.Build.Locator
         }
 
         private static string SdkResolutionExceptionMessage(string methodName) => $"Failed to find all versions of .NET Core MSBuild. Call to {methodName}. There may be more details in stderr.";
-        
+
         /// <summary>
         /// Determines the directory location of the SDK accounting for
         /// global.json and multi-level lookup policy.
@@ -256,7 +261,7 @@ namespace Microsoft.Build.Locator
             // 32-bit architecture has (x86) suffix
             string envVarName = (IntPtr.Size == 4) ? "DOTNET_ROOT(x86)" : "DOTNET_ROOT";
             var dotnetPath = FindDotnetPathFromEnvVariable(envVarName);
-            
+
             return dotnetPath;
         }
 
@@ -293,12 +298,7 @@ namespace Microsoft.Build.Locator
             string[]? resolvedPaths = null;
             foreach (string dotnetPath in s_dotnetPathCandidates.Value)
             {
-                int rc = NativeMethods.hostfxr_get_available_sdks(exe_dir: dotnetPath, result: (key, value) => resolvedPaths = value);
-
-                if (rc == 0 && resolvedPaths != null && resolvedPaths.Length > 0)
-                {
-                    break;
-                }
+                NativeMethods.hostfxr_get_available_sdks(exe_dir: dotnetPath, result: (key, value) => resolvedPaths = value);
             }
 
             // Errors are automatically printed to stderr. We should not continue to try to output anything if we failed.
@@ -321,7 +321,7 @@ namespace Microsoft.Build.Locator
         private static string? FindDotnetPathFromEnvVariable(string environmentVariable)
         {
             string? dotnetPath = Environment.GetEnvironmentVariable(environmentVariable);
-            
+
             return string.IsNullOrEmpty(dotnetPath) ? null : ValidatePath(dotnetPath);
         }
 
--- a/src/MSBuildLocator/DotNetSdkLocationHelper.cs
+++ b/src/MSBuildLocator/DotNetSdkLocationHelper.cs
@@ -72,49 +72,86 @@ namespace Microsoft.Build.Locator
 
         public static IEnumerable<VisualStudioInstance> GetInstances(string workingDirectory, bool allowQueryAllRuntimes)
         {
-            HashSet<Version> versions = new();
-            foreach (var basePath in GetDotNetBasePaths(workingDirectory))
-            {
-                var dotnetSdk = GetInstance(basePath, allowQueryAllRuntimes);
-                if (dotnetSdk != null)
-                {
-                    // Only return an SDK once, even if it's installed in multiple locations.
-                    if (versions.Add(dotnetSdk.Version))
-                    {
-                        yield return dotnetSdk;
-                    }
-                }
-            }
-        }
-
-        private static IEnumerable<string> GetDotNetBasePaths(string workingDirectory)
-        {
+            string? bestSdkPath;
+            string[] allAvailableSdks;
             try
             {
                 AddUnmanagedDllResolver();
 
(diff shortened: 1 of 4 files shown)


src/MSBuildLocator/DotNetSdkLocationHelper.cs
```diff
@@ -293,12 +298,7 @@ private static string[] GetAllAvailableSDKs()
             string[]? resolvedPaths = null;
             foreach (string dotnetPath in s_dotnetPathCandidates.Value)
             {
-                int rc = NativeMethods.hostfxr_get_available_sdks(exe_dir: dotnetPath, result: (key, value) => resolvedPaths = value);
-
-                if (rc == 0 && resolvedPaths != null && resolvedPaths.Length > 0)
-                {
-                    break;
-                }
+                NativeMethods.hostfxr_get_available_sdks(exe_dir: dotnetPath, result: (key, value) => resolvedPaths = value);
```
Forgind, line 301: How about this?

```suggestion
                int rc = NativeMethods.hostfxr_get_available_sdks(exe_dir: dotnetPath, result: (key, value) => resolvedPaths = value);

                if (rc == 0 && resolvedPaths != null && resolvedPaths.Length > 0)
                {
                    foreach (string path in resolvedPaths)
                    {
                        yield return path;
                    }
                }
```

src/MSBuildLocator/DotNetSdkLocationHelper.cs
```diff
@@ -71,45 +71,87 @@ internal static class DotNetSdkLocationHelper
         }
 
         public static IEnumerable<VisualStudioInstance> GetInstances(string workingDirectory, bool allowQueryAllRuntimes)
-        {            
-            foreach (var basePath in GetDotNetBasePaths(workingDirectory))
+        {
+            string? bestSdkPath;
+            string[] allAvailableSdks;
+            try
```
JoeRobich, line 77: Inlined `GetDotNetBasePaths`

src/MSBuildLocator/DotNetSdkLocationHelper.cs
```diff
@@ -71,45 +71,87 @@ internal static class DotNetSdkLocationHelper
...
+            foreach (var instance in instances)
             {
-                AddUnmanagedDllResolver();
+                yield return instance;
+            }
 
-                string? bestSDK = GetSdkFromGlobalSettings(workingDirectory);
-                if (!string.IsNullOrEmpty(bestSDK))
+            // Returns the list of all available SDKs ordered by ascending version.
+            static string[] GetAllAvailableSDKs()
```
JoeRobich, line 119: This method and `GetSdkFromGlobalSettings` were only referenced from the inlined `GetDotNetBasePaths` so I made them local functions.

src/MSBuildLocator/DotNetSdkLocationHelper.cs
JoeRobich: Ah, thanks in my head it was accumulating them but I see now it isn't. Will update.

JoeRobich: Updated.

src/MSBuildLocator/DotNetSdkLocationHelper.cs
```diff
@@ -71,45 +71,101 @@ internal static class DotNetSdkLocationHelper
...
 
-                string[] dotnetPaths = GetAllAvailableSDKs();
-                // We want to return the newest SDKs first, however, so iterate over the list in reverse order.
-                // If basePath is disqualified because it was later
-                // than the runtime version, this ensures that RegisterDefaults will return the latest valid
-                // SDK instead of the earliest installed.
-                for (int i = dotnetPaths.Length - 1; i >= 0; i--)
-                {
-                    if (dotnetPaths[i] != bestSDK)
+                    if (rc == 0 && resolvedPaths != null)
```
JoeRobich, line 127: Previous behavior was that we would not throw if `resolvedPaths` was an empty array.

src/MSBuildLocator/DotNetSdkLocationHelper.cs
Forgind: Definitely a bit confusing with how the native APIs work 🙂 I struggled with that for a bit when I wrote the first version of this.

src/MSBuildLocator/DotNetSdkLocationHelper.cs
```diff
@@ -71,45 +71,101 @@ internal static class DotNetSdkLocationHelper
...
+                        // We will add a null entry to the map to ensure we do not add the same SDK from a different location.
+                        versionInstanceMap[dotnetSdk.Version] = null;
+                        yield return dotnetSdk;
+                    }
+
+                    // Only add an SDK once, even if it's installed in multiple locations.
+                    if (!versionInstanceMap.ContainsKey(dotnetSdk.Version))
+                    {
+                        versionInstanceMap.Add(dotnetSdk.Version, dotnetSdk);
+                    }
```
Forgind, line 107: Tiny nit:
I think we can skip to:
```suggestion
                    // Only add an SDK once, even if it's installed in multiple locations.
                    versionInstanceMap[dotnetSdk.Version] = dotnetSdk;
```

(One fewer dictionary access)

Forgind (approved):
I think this would work the way I'd expect! @baronfel mentioned a concern about diverging from dotnet.exe's behavior in not using other dotnet.exes, but I would want to better understand his reasoning, as I'd argue that MSBuildLocator is (and always has been) more flexible with respect to which MSBuild it ultimately loads than dotnet.exe is, so I see no reason it shouldn't also be more flexible as far as which it finds.

src/MSBuildLocator/DotNetSdkLocationHelper.cs
JoeRobich: So I am unsure if it is a benefit or not, by keeping the first SDK for each discovered version we maintain a preference for SDKs in the following install location ordering.

1. DOTNET_ROOT
2. Current process path if MSBuild.Locator is called from dotnet.exe
3. DOTNET_HOST_PATH
4. DOTNET_MSBUILD_SDK_RESOLVER_CLI_DIR
5. PATH

Forgind: I'm trying to figure out if it matters or not, and I haven't come up with an answer, so playing it safe is probably a good idea.

YuliiaKovalova:
it looks good to me, @rainersigwald do you have any concerns?

src/MSBuildLocator/VisualStudioInstanceQueryOptions.cs
```diff
@@ -37,6 +37,14 @@ public class VisualStudioInstanceQueryOptions
         ///     Defaults to <see langword="false"/>. Set this to <see langword="true"/> only if your application has special logic to handle loading an incompatible SDK, such as launching a new process with the target SDK's runtime.
         /// </remarks.
         public bool AllowAllRuntimeVersions { get; set; } = false;
+
+        /// <summary>
+        ///     Allow discovery of .NET SDK versions from all discovered dotnet install locations.
+        /// </summary>
+        /// <remarks>
+        ///     Defaults to <see langword="false"/>. Set this to <see langword="true"/> only if you do not mind behaving differently than the dotnet muxer.
```
rainersigwald, line 45:
```suggestion
        ///     Defaults to <see langword="false"/>. Set this to <see langword="true"/> only if you do not mind behaving differently than a command-line dotnet invocation.
```

src/MSBuildLocator/DotNetSdkLocationHelper.cs
rainersigwald: `Dictionary<>.TryAdd()` would have that behavior right?

src/MSBuildLocator/DotNetSdkLocationHelper.cs
```diff
@@ -70,46 +70,107 @@ internal static class DotNetSdkLocationHelper
...
                 }
             }
-        }
 
-        private static IEnumerable<string> GetDotNetBasePaths(string workingDirectory)
-        {
-            try
+            // We want to return the newest SDKs first. Using OfType will remove the null entry added if we found the best SDK.
+            var instances = versionInstanceMap.Values.OfType<VisualStudioInstance>().OrderByDescending(i => i.Version);
```
rainersigwald, line 112: I am finding this sort confusing with respect to your comment above about keeping the precedence order. That is just per-version, but you'd prefer the overall results sorted by version?

Can you go into more detail on your use case?

src/MSBuildLocator/DotNetSdkLocationHelper.cs
rainersigwald: Isn't this code used only on core?

Forgind: Good point 👍

src/MSBuildLocator/DotNetSdkLocationHelper.cs
JoeRobich: I was trying to preserve the behavior from this comment:
```
// We want to return the newest SDKs first, however, so iterate over the list in reverse order.
```
But you are right. I cannot guarantee this will maintain the precedence ordering.

We intend to use this from the MSBuildWorkspace buildhost where we are looking for an SDK and we don't really care where it comes from. If they have pinned a version then we would like to find it, otherwise we would take the SDK with the highest version from any install location.

src/MSBuildLocator/DotNetSdkLocationHelper.cs
JoeRobich: I had thought about TryAdd and talked myself out of it without realizing this code only builds under .NET.

rainersigwald: In your defense the repo has "MSBuild" in the name so thinking everything targets v.OldAndBusted is not a bad assumption!

src/MSBuildLocator/DotNetSdkLocationHelper.cs
rainersigwald: Yeah that makes sense I think. The question I have is whether we should embed the sort here, or push it to the caller's layer. I don't think I have super strong opinions so this is fine.

rainersigwald (approved):
I guess this should be a minor version bump, for adding the feature to request this?

JoeRobich:
> I guess this should be a minor version bump, for adding the feature to request this?

Want me to do that in this PR? or as a follow up?

YuliiaKovalova:
> > I guess this should be a minor version bump, for adding the feature to request this?
> 
> Want me to do that in this PR? or as a follow up?

I have bumped the version , no worries!
Thank you for the contribution.

YuliiaKovalova:
It has been released:https://www.nuget.org/packages/Microsoft.Build.Locator/1.8.1 :)

PR #339, opened 2025-09-21
Modernize `DotNetSdkLocationHelper` and `hostfxr` interop.

teo-tsirpanis:
* The `NativeMethods` class was updated to use the P/Invoke source generator, and `[UnmanagedCallersOnly]`. The changes made are in a similar vein to dotnet/runtime#119034.
* The `DotNetSdkLocationHelper` class was updated to use APIs added in newer frameworks, like the regex source generator, or APIs to resolve symlinks. The latter also lets us remove the P/Invokes to `libc`.

### DIFF
--- a/src/MSBuildLocator/DotNetSdkLocationHelper.cs
+++ b/src/MSBuildLocator/DotNetSdkLocationHelper.cs
@@ -256,7 +256,7 @@ namespace Microsoft.Build.Locator
             {
                 if (!IsWindows)
                 {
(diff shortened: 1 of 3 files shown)


rainersigwald (approved):

teo-tsirpanis:
Removed assertions, as per the other native callback.

rainersigwald (approved):
Fantastic, thank you!

PR #340, opened 2025-09-26
Fix hostfxr_get_available_sdks_result_fn array size type.

tmds:
This changes the array size type to 'int' so it matches the 'int32_t' native type.

This was typed as an enum.
The enum type is accepted by CoreCLR.
Mono throws a MarshalDirectiveException because it requires the array size to be an integral type.

Fixes https://github.com/dotnet/msbuild/issues/12570.

@rainersigwald ptal

cc @jkotas @medhatiwari @giritrivedi


giritrivedi:
Can this be merged at the earliest ? S390X mono is seeing lot of failures due to this ...

PR #352, opened 2025-10-07
Add release pipeline configuration for MSBuild

YuliiaKovalova:
due to service connection requirements for symbols archival we need to migrate to classic release pipelines

### DIFF
--- a/release-pipeline.yml
+++ b/release-pipeline.yml
@@ -1,0 +1,88 @@
+trigger: none
+
+resources:
+  pipelines:
+  - pipeline: microsoft_MSBuildLocator
+    source: microsoft.MSBuildLocator
+    trigger:
+      enabled: false
+
+variables:
+  SymbolsAgentPath: '$(System.DefaultWorkingDirectory)\_$(Build.DefinitionName)\Symbols'
+  SymbolsEmailContacts: 'angerlic'
+  SymbolsFeatureName: 'MSBuild'
+  SymbolsProject: 'DDE'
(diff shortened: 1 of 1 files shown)


release-pipeline.yml
```diff
@@ -0,0 +1,88 @@
...
+resources:
+  pipelines:
+  - pipeline: microsoft_MSBuildLocator
+    source: microsoft.MSBuildLocator
+    trigger:
+      enabled: false
+
+variables:
+  SymbolsAgentPath: '$(System.DefaultWorkingDirectory)\_$(Build.DefinitionName)\Symbols'
+  SymbolsEmailContacts: 'angerlic'
```
rainersigwald, line 12:
```suggestion
  SymbolsEmailContacts: 'ykovalova'
```

release-pipeline.yml
```diff
@@ -0,0 +1,88 @@
...
+  SymbolsFeatureName: 'MSBuild'
+  SymbolsProject: 'DDE'
+  SymbolsUncPath: '\\cpvsbuild\drops\MSBuild\symbols\$(Build.DefinitionName)\$(Build.BuildNumber)\symbols.archive'
```
rainersigwald, line 15: uh, is this still how we do things? I guess it's ok as long as it works.

Maybe check in the conversion like this then remove this unused variable.

release-pipeline.yml
rainersigwald: or maybe just remove later?

rainersigwald (approved):

rainersigwald (approved):

PR #360, opened 2025-11-14
Make the MSBuild runtime assembly check more correct

baronfel:
While debugging test failures in https://github.com/CommunityToolkit/Aspire/pull/949 with @aaronpowell we discovered that the [Microsoft.SqlServer.DacFx package](https://www.nuget.org/packages/Microsoft.SqlServer.DacFx/170.2.70#dependencies-body-tab) includes a dependency on Microsoft.Build that seems to be causing MSbuild runtime dlls to end up in the test project!

This is very likely an authoring error that we should track down with them, but what surprised me was that the check for the RuntimeAssets=exclude provided by this library didn't flag this. On further investigation I discovered two gaps in the check:

* it didn't cover transitive PackageReferences to the MSBuild runtime packages from the project referencing locator
* it didn't cause the check to happen on projects that depend on the project referencing locator

The Aspire scenario is

* Package with MSBuild references (Microsoft.SqlServer.DacFx)
* is referenced by a library package that references Locator (so those MSBuild deps are transitive deps to this project, and not found)
* which is referenced by a test app project (so the check doesn't run on this project _and_ those transitive deps would not be found even if it did)

To fix this, we need to do two things
* make the check more correct by checking the runtime assets of the project in question after ResolvePackageReferences
  *  this fixes the 'transitive packageref' problem
* also ship the check as a `buildTransitive` check
  * this fixes the 'locator is referenced by a lib that is then referenced by an app' problem

Given this lib (`dotnet new classlib`):

```xml
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <TargetFramework>net10.0</TargetFramework>
    <ImplicitUsings>enable</ImplicitUsings>
    <Nullable>enable</Nullable>
  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="Microsoft.Build.Locator" Version="1.10.11-ga349ee7ffd" />
    <PackageReference Include="Microsoft.SqlServer.DacFx" Version="170.2.70" />
  </ItemGroup>

</Project>
```

and this test project referencing the lib (`dotnet new mstest`):

```xml
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <TargetFramework>net10.0</TargetFramework>
    <LangVersion>latest</LangVersion>
    <ImplicitUsings>enable</ImplicitUsings>
    <Nullable>enable</Nullable>
  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="MSTest" Version="4.0.1" />
  </ItemGroup>

  <ItemGroup>
    <Using Include="Microsoft.VisualStudio.TestTools.UnitTesting" />
  </ItemGroup>

  <ItemGroup>
    <ProjectReference Include="..\lib\lib.csproj" />
  </ItemGroup>

</Project>
```

Building the `test` project does now trigger our error:

```
> dotnet build -bl
Restore complete (0.3s)
  test net10.0 failed with 1 error(s) (0.0s)
    /Users/chethusk/.nuget/packages/microsoft.build.locator/1.10.11-ga349ee7ffd/buildTransitive/Microsoft.Build.Locator.targets(19,5): error A PackageReference to an MSBuild run-time package without ExcludeAssets="runtime" exists in your project. This will cause MSBuild run-time assemblies to be copied to your output directory, causing your application to load them at runtime. To use the copy of MSBuild registered by MSBuildLocator, set ExcludeAssets="runtime" on the MSBuild PackageReferences. To disable this check, set the property DisableMSBuildAssemblyCopyCheck=true in your project file (not recommended as you must distributed all of MSBuild + associated toolset). Package(s) referenced: Microsoft.Build@15.9.20, Microsoft.Build.Framework@15.9.20, Microsoft.Build.Utilities.Core@15.9.20
```

While in here, I also added a small README for the package so that the NuGet.org experience is nicer.

### DIFF
--- a/src/MSBuildLocator/build/Microsoft.Build.Locator.targets
+++ b/src/MSBuildLocator/build/Microsoft.Build.Locator.targets
@@ -14,11 +14,12 @@
       <_MSBuildAssembliesCopyLocalItems Include="@(RuntimeCopyLocalItems->WithMetadataValue('NuGetPackageId', 'NuGet.Frameworks'))" />
       <MSBuildAssembliesCopyLocalItems Include="@(_MSBuildAssembliesCopyLocalItems->WithMetadataValue('CopyLocal', 'true')->WithMetadataValue('AssetType', 'runtime'))" />
       <_DistinctMSBuildPackagesReferencedPoorly Include="@(MSBuildAssembliesCopyLocalItems->Metadata('NuGetPackageId')->Distinct())" />
-      <_DistinctPackagesWithVersions Include="@(_DistinctMSBuildPackagesReferencedPoorly->'%(NuGetPackageId)@%(NuGetPackageVersion)')" />
     </ItemGroup>
 
     <Error
-      Condition="@(_DistinctPackagesWithVersions->Count()) > 0"
-      Text="A PackageReference to an MSBuild run-time package without ExcludeAssets=&quot;runtime&quot; exists in your project. This will cause MSBuild run-time assemblies to be copied to your output directory, causing your application to load them at runtime. To use the copy of MSBuild registered by MSBuildLocator, set ExcludeAssets=&quot;runtime&quot; on the MSBuild PackageReferences. To disable this check, set the property DisableMSBuildAssemblyCopyCheck=true in your project file (not recommended as you must distributed all of MSBuild + associated toolset). Package(s) referenced: @(_DistinctPackagesWithVersions, ', ')" />
+      Code="MSBL001"
+      Text="A PackageReference to the package '%(_DistinctMSBuildPackagesReferencedPoorly.NuGetPackageId)' at version '%(_DistinctMSBuildPackagesReferencedPoorly.NuGetPackageVersion)' is present in this project without ExcludeAssets=&quot;runtime&quot; set. This can cause errors at run-time due to MSBuild assembly-loading."
+      HelpLink="https://aka.ms/msbuild/locator/diagnostics/MSBL001"
+      />
   </Target>
 </Project>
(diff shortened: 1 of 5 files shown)


src/MSBuildLocator/build/Microsoft.Build.Locator.targets
```diff
@@ -1,25 +1,24 @@
...
+      <_DistinctMSBuildPackagesReferencedPoorly Include="@(MSBuildAssembliesCopyLocalItems->Metadata('NuGetPackageId')->Distinct())" />
+      <_DistinctPackagesWithVersions Include="@(_DistinctMSBuildPackagesReferencedPoorly->'%(NuGetPackageId)@%(NuGetPackageVersion)')" />
     </ItemGroup>
+
     <Error
-      Condition="'@(MSBuildPackagesWithoutPrivateAssets)' != ''"
-      Text="A PackageReference to Microsoft.Build.* without ExcludeAssets=&quot;runtime&quot; exists in your project. This will cause MSBuild assemblies to be copied to your output directory, causing your application to load them at runtime. To use the copy of MSBuild registered by MSBuildLocator, set ExcludeAssets=&quot;runtime&quot; on the MSBuild PackageReferences. To disable this check, set the property DisableMSBuildAssemblyCopyCheck=true in your project file (not recommended as you must distributed all of MSBuild + associated toolset). Package(s) referenced: @(MSBuildPackagesWithoutPrivateAssets)" />
+      Condition="@(_DistinctPackagesWithVersions->Count()) > 0"
```
baronfel: I kind of hate how long this message is, and I would like to consider issueing one error message per package violation instead.

baronfel: Ok, I've made the diagnostic trigger once per package violation now. The error now looks like:

```
➜  test dotnet build
  test net10.0 failed with 1 error(s) (0.1s)
    /Users/chethusk/.nuget/packages/microsoft.build.locator/1.10.12-gf43ae4a423/buildTransitive/Microsoft.Build.Locator.targets(19,5): error MSBL001: A PackageReference to the package 'Microsoft.Build' at version '15.9.20' is present in this project without ExcludeAssets="runtime" set. This can cause errors at run-time due to MSBuild assembly-loading.
```

The diagnostic code embeds the aka.ms url https://aka.ms/msbuild/locator/diagnostics/MSBL001, which I will set up to point to the new diagnostic doc in this PR.

YuliiaKovalova (approved):

rainersigwald (approved):

Issue #282, opened 2024-02-26
List of Breaking changes for MSBuild.Locator v 2.0


YuliiaKovalova:
1. Remove extra property AllowQueryAllRuntimeVersions. AllowAllRuntimeVersions from [VisualStudioInstanceQueryOptions.cs](https://github.com/microsoft/MSBuildLocator/pull/281/files/ca50741dbefae60fca50c6098f196831e75cc798#diff-1f6b7466885c2f07527b05f53408d46752261ab86dfbe34127493417d6f60474) means the same, but can't clean up immediately due to backward compatibility.

rainersigwald:
After reading what you wrote above, I added a `Version` field to the result, which may actually supersede some of the warning stuff.

rainersigwald:
We should also rename all the core types so they don't talk about VisualStudio on core.

PR #381, opened 2026-06-04
Add AGENTS.md and per-TFM MSBuild loader skills

rainersigwald:
## Why

MSBuildLocator's implementation forks heavily between `net46` (Visual Studio) and `net8.0` (.NET SDK). This PR adds Copilot agent instructions and splits the documentation into an always-loaded base file and two on-demand skills to prevent context bloat and confusion between the frameworks.

## What's added

- **`AGENTS.md`**: Core repository context, build/test commands, conventions, and the cross-TFM register-before-load contract.
- **`msbuild-loader-netcore` skill**: `net8.0` / `NETCOREAPP` specifics (`AssemblyLoadContext`, hostfxr SDK discovery, probe order, SDK environment variables).
- **`msbuild-loader-netframework` skill**: `net46` specifics (`AppDomain.AssemblyResolve`, Developer Console / VS Setup COM discovery).

## Notes

- Content was LLM-verified against the source code and [official MSBuild documentation](https://learn.microsoft.com/en-us/visualstudio/msbuild/find-and-use-msbuild-versions).
- Documentation only; no code changes.
- I did `AGENTS.md` pretty manually but the skills are generated.

### DIFF
--- a/.agents/skills/msbuild-loader-netcore/SKILL.md
+++ b/.agents/skills/msbuild-loader-netcore/SKILL.md
@@ -1,0 +1,80 @@
+---
+name: msbuild-loader-netcore
+description: >-
+  How MSBuildLocator loads MSBuild assemblies and discovers the .NET SDK on .NET
+  / .NET Core (the net8.0 target / #if NETCOREAPP branches in src/MSBuildLocator).
+  Covers the AssemblyLoadContext.Default.Resolving handler, search-path probing,
+  the SDK environment variables set on registration, hostfxr-based SDK discovery,
+  the dotnet location probe order, and the AllowQueryAll* widening flags. Use when
+  editing or reviewing net8.0 loader, registration, or SDK-discovery code, or
+  diagnosing assembly-resolution behavior on .NET Core hosts.
+---
+
+# .NET / .NET Core (`net8.0` / `NETCOREAPP`) MSBuild loader
+
+Scope: `src/MSBuildLocator/MSBuildLocator.cs` `#if NETCOREAPP` branches, plus
+`DotNetSdkLocationHelper.cs` and `NativeMethods.cs`. For build/test and
+cross-cutting conventions, see `AGENTS.md`.
+
+## Assembly-resolution handler
+- `s_registeredHandler` is a static
+  `Func<AssemblyLoadContext, AssemblyName, Assembly>`; registration hooks
+  `AssemblyLoadContext.Default.Resolving`.
+- Resolving can fire repeatedly; successful loads are cached in the local
+  `loadedAssemblies` dictionary keyed by `AssemblyName.FullName`.
+- Resolution is not thread-safe; keep the `loadedAssemblies` lock around cache
+  lookup, path probing, `Assembly.LoadFrom`, and cache insert.
+- The resolver receives `AssemblyName` directly — do not parse an event-args
+  name string (that is the net46 path).
+- For each registered `msbuildSearchPaths` entry, probe
+  `Path.Combine(msbuildPath, assemblyName.Name + ".dll")` and load with
+  `Assembly.LoadFrom(targetAssembly)`.
+
+## SDK environment variables (NETCOREAPP only)
+- Registration sets MSBuild SDK env vars via `ApplyDotNetSdkEnvironmentVariables`:
+  - `MSBUILD_EXE_PATH` = `<sdk>\MSBuild.dll`
+  - `MSBuildExtensionsPath` = `<sdk>`
+  - `MSBuildSDKsPath` = `<sdk>\Sdks`
+- `RegisterMSBuildPath(string)` applies these for that path before registering.
+- `RegisterMSBuildPath(string[])` applies these for the first search path only,
+  then registers all paths.
+- `RegisterInstance` applies these only when
+  `instance.DiscoveryType == DiscoveryType.DotNetSdk`.
+- Do not move this setup into net46; Framework has its own `MSBUILD_EXE_PATH`
+  compatibility branch under `#if NET46` (see the `msbuild-loader-netframework`
+  skill).
+
+## SDK discovery
+- Lives in `DotNetSdkLocationHelper`; instances are
+  `VisualStudioInstance(name: ".NET Core SDK", ..., DiscoveryType.DotNetSdk)`.
+- Uses `NativeMethods` hostfxr P/Invoke under `#if NETCOREAPP` only:
+  - `hostfxr_resolve_sdk2` — best SDK, honoring `global.json` via
+    `WorkingDirectory`.
+  - `hostfxr_get_available_sdks` — installed SDK enumeration.
+- `ResolveDotnetPathCandidates` preference order (tried in order):
+  `DOTNET_ROOT` (`DOTNET_ROOT(x86)` in a 32-bit process) → current process
+  directory when running under `dotnet` → `DOTNET_HOST_PATH` →
+  `DOTNET_MSBUILD_SDK_RESOLVER_CLI_DIR` → `PATH`.
+- A successful `hostfxr_resolve_sdk2` sets `DOTNET_HOST_PATH` (if empty) to
+  `<dotnetPath>\dotnet(.exe)`.
+- Default query returns the best SDK first, then unique SDK versions
+  newest-first from the available SDKs.
+
+## Widening flags
+- `AllowQueryAllRuntimeVersions` / `VisualStudioInstanceQueryOptions.AllowAllRuntimeVersions`
+  include SDKs whose major/minor runtime exceeds `Environment.Version`.
+- `AllowQueryAllDotnetLocations` / `VisualStudioInstanceQueryOptions.AllowAllDotnetLocations`
+  keep probing all dotnet candidate locations instead of stopping after the
+  first location that has SDKs.
+
+## What net8.0 does NOT do
+- No Developer Console or Visual Studio Setup COM discovery;
+  `FEATURE_VISUALSTUDIOSETUP` package references/constants are net46-only in the
+  csproj.
+
+## Register-before-load contract
+- `CanRegister` is false when already registered, or once any signed
+  `Microsoft.Build*` core assembly is loaded.
+- JIT caveat: JIT-compilation of a method referencing `Microsoft.Build` types is
+  enough to load those assemblies and break registration. Keep locator calls
+  isolated before any such reference.
--- a/.agents/skills/msbuild-loader-netcore/SKILL.md
+++ b/.agents/skills/msbuild-loader-netcore/SKILL.md
@@ -16,6 +16,8 @@ Scope: `src/MSBuildLocator/MSBuildLocator.cs` `#if NETCOREAPP` branches, plus
 `DotNetSdkLocationHelper.cs` and `NativeMethods.cs`. For build/test and
 cross-cutting conventions, see `AGENTS.md`.
 
+No platform-specific code should be required for .NET SDK discovery.
+
 ## Assembly-resolution handler
 - `s_registeredHandler` is a static
   `Func<AssemblyLoadContext, AssemblyName, Assembly>`; registration hooks
--- a/.agents/skills/msbuild-loader-netcore/SKILL.md
+++ b/.agents/skills/msbuild-loader-netcore/SKILL.md
@@ -16,7 +16,7 @@ Scope: `src/MSBuildLocator/MSBuildLocator.cs` `#if NETCOREAPP` branches, plus
 `DotNetSdkLocationHelper.cs` and `NativeMethods.cs`. For build/test and
 cross-cutting conventions, see `AGENTS.md`.
 
-SDK discovery is cross-platform, but executable names, hostfxr library loading, and symlink handling have OS-specific implementations.```
+SDK discovery is cross-platform, but executable names, hostfxr library loading, and symlink handling have OS-specific implementations.
 
 ## Assembly-resolution handler
 - `s_registeredHandler` is a static
--- a/.agents/skills/msbuild-loader-netcore/SKILL.md
+++ b/.agents/skills/msbuild-loader-netcore/SKILL.md
@@ -16,7 +16,7 @@ Scope: `src/MSBuildLocator/MSBuildLocator.cs` `#if NETCOREAPP` branches, plus
 `DotNetSdkLocationHelper.cs` and `NativeMethods.cs`. For build/test and
 cross-cutting conventions, see `AGENTS.md`.
 
-No platform-specific code should be required for .NET SDK discovery.
+SDK discovery is cross-platform, but executable names, hostfxr library loading, and symlink handling have OS-specific implementations.```
 
 ## Assembly-resolution handler
 - `s_registeredHandler` is a static
--- a/.agents/skills/msbuild-loader-netframework/SKILL.md
+++ b/.agents/skills/msbuild-loader-netframework/SKILL.md
@@ -1,0 +1,67 @@
+---
+name: msbuild-loader-netframework
+description: >-
+  How MSBuildLocator loads MSBuild assemblies and discovers installs on .NET
+  Framework (the net46 target / #if NET46 / non-NETCOREAPP branches in
+  src/MSBuildLocator). Covers the AppDomain.AssemblyResolve handler, search-path
+  probing, the pre-17.1 MSBuild.exe x86/amd64 fix, and Developer Console + Visual
+  Studio Setup (COM) discovery. Use when editing or reviewing net46 loader,
+  registration, or VS-discovery code, or diagnosing assembly-resolution behavior
+  on .NET Framework hosts.
+---
+
+# .NET Framework (`net46`) MSBuild loader
+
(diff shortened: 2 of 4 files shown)


JanProvaznik (approved):

.agents/skills/msbuild-loader-netframework/SKILL.md
```diff
@@ -0,0 +1,67 @@
...
+  `VisualStudioLocationHelper.GetInstances()` enumerates VS 2017+ setup instances
+  with `Microsoft.Component.MSBuild`; yields `DiscoveryType.VisualStudioSetup`.
+- `DiscoveryType.DotNetSdk` exists in the enum but belongs to the Core path; net46
+  `GetInstances(...)` does not call SDK discovery.
+
+## What net46 does NOT do
+- No `AssemblyLoadContext`, `hostfxr`, or `.NET SDK` discovery — those are
+  `#if NETCOREAPP` paths (see the `msbuild-loader-netcore` skill).
+- `ApplyDotNetSdkEnvironmentVariables(...)` exists in the file, but both
+  `RegisterMSBuildPath(...)` call sites that invoke it are `#if NETCOREAPP`. Do
```
OvesN: Actually we have three ApplyDotNetSdkEnvironmentVariables calls, and this third one not mentioned here is not if-gated but is guarded at runtime by DiscoveryType == DotNetSdk

.agents/skills/msbuild-loader-netcore/SKILL.md
```diff
@@ -0,0 +1,80 @@
...
+  / .NET Core (the net8.0 target / #if NETCOREAPP branches in src/MSBuildLocator).
+  Covers the AssemblyLoadContext.Default.Resolving handler, search-path probing,
+  the SDK environment variables set on registration, hostfxr-based SDK discovery,
+  the dotnet location probe order, and the AllowQueryAll* widening flags. Use when
+  editing or reviewing net8.0 loader, registration, or SDK-discovery code, or
+  diagnosing assembly-resolution behavior on .NET Core hosts.
+---
+
+# .NET / .NET Core (`net8.0` / `NETCOREAPP`) MSBuild loader
+
```
OvesN, line 14: Would be great to have here non-window specific logic listed, how the flow differs for windows vs other OS

AGENTS.md
```diff
@@ -0,0 +1,31 @@
+# Microsoft.Build.Locator — Copilot instructions
```
OvesN, line 1: Is an agent smart enough to know what if we change logic described in skills agent should change skill as well?

.agents/skills/msbuild-loader-netcore/SKILL.md
```diff
@@ -0,0 +1,80 @@
...
+  then registers all paths.
+- `RegisterInstance` applies these only when
+  `instance.DiscoveryType == DiscoveryType.DotNetSdk`.
+- Do not move this setup into net46; Framework has its own `MSBUILD_EXE_PATH`
+  compatibility branch under `#if NET46` (see the `msbuild-loader-netframework`
+  skill).
+
+## SDK discovery
+- Lives in `DotNetSdkLocationHelper`; instances are
+  `VisualStudioInstance(name: ".NET Core SDK", ..., DiscoveryType.DotNetSdk)`.
```
OvesN, line 51: I am confused, why this is called VisualStudioInstance but we have here DiscoveryType.DotNetSdk?
Does VisualStudioInstance really represent installed visual studio or installed msbuild?

AGENTS.md
```diff
@@ -0,0 +1,31 @@
...
+
+## Architecture (namespace `Microsoft.Build.Locator`)
+- `MSBuildLocator.cs` — entry point: `RegisterDefaults`, `RegisterInstance`, `RegisterMSBuildPath`, `QueryVisualStudioInstances`, `CanRegister`, handler registration. Both TFM forks live here. `Unregister()` is kept only for back-compat and is a no-op (the resolver is never removed).
+- `DotNetSdkLocationHelper.cs`, `NativeMethods.cs` — `.NET SDK` discovery (hostfxr); see the `msbuild-loader-netcore` skill.
+- `VisualStudioLocationHelper.cs` — Visual Studio Setup (COM) discovery; see the `msbuild-loader-netframework` skill.
+- `VisualStudioInstance.cs`, `VisualStudioInstanceQueryOptions.cs`, `DiscoveryType.cs` — result/option types.
+- `Utils/SemanticVersion*.cs`, `VersionComparer.cs` — internal SemVer parse/compare to order instances.
+- Props/targets shipped from `src/MSBuildLocator/build/` (packed to `build/` + `buildTransitive/`). `EnsureMSBuildAssembliesNotCopied` emits error **MSBL001** when a consumer references `Microsoft.Build*` packages without `PrivateAssets="all"` / `ExcludeAssets="runtime"` (copying those assemblies locally breaks redirection). Keep the target's package list in sync with MSBuild's layout.
```
OvesN: This point seems a little confusing if a human will read this. (at least for me)
I suggest to write it in simple words and into two different bullet points:

1) The core rule of MSBuildLocator: never ship MSBuild's DLLs with the app.
 The app should ship without  Microsoft.Build.dll . At runtime, Locator's registered handler redirects those loads to the real MSBuild  install (VS or SDK).

 • Why a local copy breaks it. The resolve handler only fires on a miss — when the loader can't find an assembly. If a copy of     Microsoft.Build.dll  sits in the  bin /output folder, the loader finds that one and the handler never runs.
 • MSBL001 safeguard. The Locator NuGet package ships an MSBuild  .targets  file that runs inside the consuming build. The     EnsureMSBuildAssembliesNotCopied  target scans the resolved packages for any MSBuild package about to be copied locally. If it    finds one, the build fails with error MSBL001, which names the offending package and asks for two attributes.
 • The fix — two attributes, on the flagged  <PackageReference>  in the consuming project's  .csproj . Add  ExcludeAssets="runtime"     (stops the DLL from being copied to output) and  PrivateAssets="all"  (keeps the dependency from flowing to downstream consumers):
   <PackageReference Include="Microsoft.Build" Version="…"
                     ExcludeAssets="runtime" PrivateAssets="all" />

 2) Keep the target's package list in sync.
 The target checks a hardcoded list of package IDs ( Microsoft.Build ,  .Framework ,  .Utilities.Core ,  .Tasks.Core , …) — see   Microsoft.Build.Locator.targets#L3-L14 . If MSBuild ever ships a new redistributable assembly that also shouldn't be copied, it must  be added here or the guard won't catch it.

.agents/skills/msbuild-loader-netcore/SKILL.md
```diff
@@ -0,0 +1,80 @@
...
+  `loadedAssemblies` dictionary keyed by `AssemblyName.FullName`.
+- Resolution is not thread-safe; keep the `loadedAssemblies` lock around cache
+  lookup, path probing, `Assembly.LoadFrom`, and cache insert.
+- The resolver receives `AssemblyName` directly — do not parse an event-args
+  name string (that is the net46 path).
+- For each registered `msbuildSearchPaths` entry, probe
+  `Path.Combine(msbuildPath, assemblyName.Name + ".dll")` and load with
+  `Assembly.LoadFrom(targetAssembly)`.
+
+## SDK environment variables (NETCOREAPP only)
```
OvesN, line 35: Would be great to have here some context why this env variables are important to msbuildLocator

.agents/skills/msbuild-loader-netframework/SKILL.md
```diff
@@ -0,0 +1,67 @@
...
+  `lock (loadedAssemblies)`.
+- Handler path: parse `eventArgs.Name` with `new AssemblyName(eventArgs.Name)`;
+  for each registered search path, if `<msbuildPath>\<Name>.dll` exists, return
+  `Assembly.LoadFrom(targetAssembly)`.
+- Search paths come from `RegisterMSBuildPath(...)`, or from
+  `RegisterInstance(...)` as `instance.MSBuildPath` plus the VS NuGet path when
+  it exists.
+
+## Pre-17.1 MSBuild.exe bitness fix
+- net46 reads `FileVersionInfo` from the first registered path containing
```
OvesN: Mb add here why this bug exists

OvesN (approved):
If we want to treat SKILL.md also as a documentation, I would expand it a little bit.

.agents/skills/msbuild-loader-netcore/SKILL.md
rainersigwald: There shouldn't be a difference beyond there not being a net472 option off of Windows.

.agents/skills/msbuild-loader-netcore/SKILL.md
rainersigwald: It would! I _think_ we shouldn't need these any more and just finding the right SDK should do the trick. But I'm not proposing changing that right now

.agents/skills/msbuild-loader-netcore/SKILL.md
rainersigwald: It's misnamed. It should be InstalledMSBuild or something.

(We picked the name before supporting finding the SDK)

rainersigwald: https://github.com/microsoft/MSBuildLocator/issues/282#issuecomment-4773359318

AGENTS.md
rainersigwald: @copilot make this fix

.agents/skills/msbuild-loader-netframework/SKILL.md
rainersigwald: Yeah this is one it backed out from the code and it should probably just be dropped

.agents/skills/msbuild-loader-netcore/SKILL.md
rainersigwald: Added as https://github.com/microsoft/MSBuildLocator/pull/381/changes#diff-597e7c68ae23dccefb15767cf3ee591ee6ad0fd17378fe00eda12054fb0d8f12R19.

.agents/skills/msbuild-loader-netcore/SKILL.md
rainersigwald: #389

.agents/skills/msbuild-loader-netframework/SKILL.md
rainersigwald: You're right; I'm going to just remove this section since we're likely to drop those environment variables soon.

.agents/skills/msbuild-loader-netframework/SKILL.md
```diff
@@ -0,0 +1,56 @@
...
+of `NETCOREAPP`) branches. `FEATURE_VISUALSTUDIOSETUP` is defined only when
+`TargetFramework == net46` in `Microsoft.Build.Locator.csproj`. For build/test
+and cross-cutting conventions, see `AGENTS.md`.
+
+## Assembly-resolution handler
+- `s_registeredHandler` is a static `ResolveEventHandler`; `IsRegistered` is
+  `s_registeredHandler != null`.
+- `RegisterMSBuildPathsInternally` stores the handler in the static field before
+  subscribing to `AppDomain.CurrentDomain.AssemblyResolve`; the static field
+  keeps the delegate alive so it persists.
```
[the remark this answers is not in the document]
rainersigwald: Accepted. I updated the skill to clarify that the event subscription keeps the delegate alive, while the static field tracks registration state through IsRegistered.

.agents/skills/msbuild-loader-netcore/SKILL.md
```diff
@@ -0,0 +1,82 @@
...
+  diagnosing assembly-resolution behavior on .NET Core hosts.
+---
+
+# .NET / .NET Core (`net8.0` / `NETCOREAPP`) MSBuild loader
+
+Scope: `src/MSBuildLocator/MSBuildLocator.cs` `#if NETCOREAPP` branches, plus
+`DotNetSdkLocationHelper.cs` and `NativeMethods.cs`. For build/test and
+cross-cutting conventions, see `AGENTS.md`.
+
+No platform-specific code should be required for .NET SDK discovery.
```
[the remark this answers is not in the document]
rainersigwald:
```suggestion
SDK discovery is cross-platform, but executable names, hostfxr library loading, and symlink handling have OS-specific implementations.```

File: AGENTS.md (hunk 3 above)
Comment (rainersigwald) [reply AGENTS.md:1]: added "When changing behavior documented by either skill, update the skill in the same change."

File: AGENTS.md [hunk 10]
```diff
@@ -0,0 +1,34 @@
...
+Library: `net46` + `net8.0`. Tests: `net472` + `net8.0`. Non-trivial code forks per TFM via `#if NETCOREAPP`, `#if NET46`, and `FEATURE_VISUALSTUDIOSETUP` (defined only for `net46`). Always check whether a change must be mirrored or excluded across these conditionals; for what each fork actually does, load the `msbuild-loader-netcore` or `msbuild-loader-netframework` skill.
+
+When changing behavior documented by either skill, update the skill in the same change.
+
+## Architecture (namespace `Microsoft.Build.Locator`)
+- `MSBuildLocator.cs` — entry point and handler registration. Both TFM forks live here. `Unregister()` is retained for compatibility but is a no-op.
+- `DotNetSdkLocationHelper.cs`, `NativeMethods.cs` — .NET SDK discovery (hostfxr); see `msbuild-loader-netcore`.
+- `VisualStudioLocationHelper.cs` — `net46`-only Visual Studio Setup discovery; see `msbuild-loader-netframework`.
+- `VisualStudioInstance.cs`, `VisualStudioInstanceQueryOptions.cs`, `DiscoveryType.cs` — result/option types.
+- `Utils/SemanticVersion*.cs`, `VersionComparer.cs` — internal SemVer parse/compare to order instances.
```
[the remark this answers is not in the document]
rainersigwald:
```suggestion
- `Utils/SemanticVersion*.cs`, `VersionComparer.cs` — internal SemVer parse/compare that is an implementation detail of .NET SDK discovery.
```

AGENTS.md
```diff
@@ -0,0 +1,34 @@
...
+- Keep `EnsureMSBuildAssembliesNotCopied`'s hardcoded package list synchronized with MSBuild's redistributable assemblies so it catches new packages.
+
+## Conventions
+- Contract-stable public API: csproj `EnablePackageValidation` + `PackageValidationBaselineVersion` (1.6.1). Intentional API changes require updating `src/MSBuildLocator/CompatibilitySuppressions.xml` and an appropriate semver update.
+- XML doc comments on public members; match existing style.
+- Strong-name signed (`key.snk`) — don't remove signing.
+- Build settings centralized in `Directory.Build.props` / `Directory.Solution.props` / `Directory.Build.rsp` — edit there, not per-project.
+- Register-before-load contract: callers must register via Locator BEFORE any `Microsoft.Build.*` type loads (`CanRegister` → false once loaded). Preserve this + the lazy-loading patterns protecting it when refactoring.
```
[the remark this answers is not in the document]
rainersigwald:
```suggestion
 - Register-before-load contract: callers must register via Locator BEFORE any core MSBuild assembly loads (`CanRegister` → false once loaded). Preserve this + the lazy-loading patterns protecting it when refactoring.```

File: .agents/skills/msbuild-loader-netcore/SKILL.md [hunk 12]
```diff
@@ -0,0 +1,82 @@
...
+  diagnosing assembly-resolution behavior on .NET Core hosts.
+---
+
+# .NET / .NET Core (`net8.0` / `NETCOREAPP`) MSBuild loader
+
+Scope: `src/MSBuildLocator/MSBuildLocator.cs` `#if NETCOREAPP` branches, plus
+`DotNetSdkLocationHelper.cs` and `NativeMethods.cs`. For build/test and
+cross-cutting conventions, see `AGENTS.md`.
+
+SDK discovery is cross-platform, but executable names, hostfxr library loading, and symlink handling have OS-specific implementations.```
```
OvesN: nit: stray ```

AGENTS.md
```diff
@@ -0,0 +1,34 @@
...
+- Keep `EnsureMSBuildAssembliesNotCopied`'s hardcoded package list synchronized with MSBuild's redistributable assemblies so it catches new packages.
+
+## Conventions
+- Contract-stable public API: csproj `EnablePackageValidation` + `PackageValidationBaselineVersion` (1.6.1). Intentional API changes require updating `src/MSBuildLocator/CompatibilitySuppressions.xml` and an appropriate semver update.
+- XML doc comments on public members; match existing style.
+- Strong-name signed (`key.snk`) — don't remove signing.
+- Build settings centralized in `Directory.Build.props` / `Directory.Solution.props` / `Directory.Build.rsp` — edit there, not per-project.
+ - Register-before-load contract: callers must register via Locator BEFORE any core MSBuild assembly loads (`CanRegister` → false once loaded). Preserve this + the lazy-loading patterns protecting it when refactoring.```
```
OvesN: nit: stray ```

OvesN (approved):

OvesN (approved):

PR #384, opened 2026-06-04
Avoid unnecessary restore when running tests in PR builds

rainersigwald:
Honestly `dotnet test --no-build` should imply `--no-restore` . . .


rainersigwald:
Looks like this doesn't work and I'm not going to figure out why right now.

